PR #10, opened 2025-12-25
Added background sender validation

denislupookov:
[Ticket ID](url)

### Requirements

Message handlers don't validate `sender.url` or `sender.tab`, so any malicious browser extensions can send fake messages to steal data.

### Changes

Added a check so unauthorized messages return `Unauthorized` instead of executing sensitive operations.

### Test

I added test coverage for the function to see the result.
Also this 2 screenshots show that the check works if we decide to send the request inside the browser extension to our content-script without tab id.

Before: 
Screenshot 2025-12-26 at 13 41 26

After:
Screenshot 2025-12-26 at 13 41 04

### DIFF
--- a/src/background/index.js
+++ b/src/background/index.js
@@ -7,6 +7,7 @@ import { arrayBufferToBase64Url } from '../shared/utils/arrayBufferToBase64Url'
 import { base64UrlToArrayBuffer } from '../shared/utils/base64UrlToArrayBuffer'
 import { logger } from '../shared/utils/logger'
 import { validateSender } from './utils/validateSender'
+import { MESSAGE_TYPES, SECURE_MESSAGE_TYPES } from '../shared/services/messageBridge'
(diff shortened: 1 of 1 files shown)


src/background/index.js
```diff
@@ -69,6 +70,30 @@ chrome.alarms.onAlarm.addListener(async (alarm) => {
 })
 
 chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
+
+  const sensitiveTypes = [
```
giorgikh93: We have message types created here - https://github.com/tetherto/pearpass-app-browser-extension/blob/main/src/shared/services/messageBridge.js#L15, wdyt? is it the one we should be using here instead of creating new types?

giorgikh93 (approved):

eugeneglova (approved):

PR #11, opened 2025-12-25
Add msg and event origin validation for password suggestions.

harrisaar-noxtton:
Potential Attack Scenario (PoC Summary)

1. A malicious website locates PearPass-related iframes using querySelector.
2. The malicious site sends crafted postMessage events to the iframe with a spoofed origin.
3. The iframe does not properly validate the message origin.
4. The extension displays incorrect domain suggestions.
5. The malicious website overlays fake but realistic UI elements (clickjacking).
6. The user unknowingly triggers autofill for a different domain.
7. Autofilled credentials could be exposed or recorded without user awareness.

---
  - https://app.asana.com/0/0/1212541757842133

### DIFF
--- a/src/contentPopups/utils/messageValidation.js
+++ b/src/contentPopups/utils/messageValidation.js
@@ -1,24 +1,38 @@
+import { logger } from '../../shared/utils/logger'
 
-export const isMessageOriginValid = (msg, eventOrigin) => {
-    if (!msg?.data?.url) {
-        return true
-    }
+/**
+ * Validates that the origin of the message event matches the origin of the URL provided in the message payload.
+ *
+ * @param {Object} msg - The message object received from the postMessage event.
+ * @param {Object} msg.data - The data payload of the message.
+ * @param {string} msg.data.url - The URL of the page sending the message. Required string, but monitored for null values.
+ * @param {string} eventOrigin - The `origin` property of the MessageEvent. Expected to be a non-null string.
+ * @returns {boolean} Returns `true` if the origins match, otherwise `false`.
+ */
+export const doesPayloadUrlMatchOrigin = (msg, eventOrigin) => {
+  if (!msg?.data?.url) {
+    return false
+  }
 
-    try {
-        const urlOrigin = new URL(msg.data.url).origin
-        if (urlOrigin !== eventOrigin) {
-            console.warn(
-                'Security Warning: Message origin does not match URL origin',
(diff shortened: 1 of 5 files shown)


torkinos (approved):

src/contentPopups/utils/messageValidation.js
```diff
@@ -0,0 +1,24 @@
+
```
eugeneglova: extra empty line

eugeneglova (approved):

src/contentPopups/utils/messageValidation.js
```diff
@@ -0,0 +1,24 @@
...
+export const isMessageOriginValid = (msg, eventOrigin) => {
+    if (!msg?.data?.url) {
+        return true
```
alvaropaco: Think this could be a potential security issue:. This is dangerous in a postMessage or cross-window communication context. An attacker can simply omit the url field and bypass validation entirely.

```js
if (!msg?.data?.url) {
    return false
}
```

src/contentPopups/utils/messageValidation.test.js
```diff
@@ -0,0 +1,71 @@
...
+        const msg = {
+            data: {
+                url: 'not-a-valid-url'
+            }
+        }
+        const origin = 'https://example.com'
+        expect(isMessageOriginValid(msg, origin)).toBe(false)
+        expect(console.error).toHaveBeenCalled()
+    })
+})
```
alvaropaco:
```suggestion
})

```

src/contentPopups/utils/messageValidation.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    try {
+        const urlOrigin = new URL(msg.data.url).origin
```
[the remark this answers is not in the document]
harrisaar-noxtton: Maybe enforcement should be at a higher level. The PearPass extension should throw a big error if the user is on an HTTP page. If we check the URL in every function, it’s duplication. PearPass might already be checking the HTTPS protocol—I need to verify.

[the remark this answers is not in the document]
torkinos: As we discussed, I would suggest to resolve this comment and address this issue via separate PR (Implementing Opt-in for HTTP pages)

torkinos (approved):

PR #21, opened 2026-01-06
Passkey private key storage

shavtvalishvili:
### Requirements

- Passkey private keys no longer stored as a PEM string with easily identifiable headers and raw private key string.

### Changes

- Store passkey private key as raw binary PKCS#8 buffer encoded as Base64URL.
- Fix passkey input field not being readonly in CreateOrEditLogin view.

### Testing Notes

- Create a passkey and authenticate using it on a passkey-supported website like google.com, github.com.

---
  - https://app.asana.com/0/0/1212541758861740

### DIFF
--- a/src/action/containers/CreateOrEditLogin/index.jsx
+++ b/src/action/containers/CreateOrEditLogin/index.jsx
@@ -186,7 +186,7 @@ export const CreateOrEditLogin = ({
               placeholder={t`Passkey Stored`}
               variant="outline"
               icon={KeyIcon}
-              readonly={true}
+              readonly
             />
           )}
         </FormGroup>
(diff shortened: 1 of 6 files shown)


src/action/containers/CreateOrEditLogin/index.jsx
```diff
@@ -186,7 +186,7 @@ export const CreateOrEditLogin = ({
               placeholder={t`Passkey Stored`}
               variant="outline"
               icon={KeyIcon}
-              disabled
+              readonly={true}
```
boris91: redundant `={true}`?

boris91 (approved):

PR #32, opened 2026-01-15
Native messaging handshake vulnerability

shavtvalishvili:
### Changes

- Added a client key store that creates and saves a long‑term Ed25519 key pair, encrypted with the master password and stored in IndexedDB.
- When the user enters their master password, the client key pair is decrypted and kept only in memory for the current session.
- Updated the secure channel handshake to include the client public key in the transcript and sign the transcript with the client private key.
- Added a background command to unlock the client key store without dropping the current session.
- Adjusted error handling so master‑password and client‑signature errors no longer automatically clear pairing.
- Added shared constants for new auth/native‑messaging error types and shared base64 helpers.
- Expanded unit tests for secure channel and native messaging to cover the new client identity, unlock flow, and non‑destructive error behavior.

### Testing Notes

- Smoke tested extension operation.
- Tested desktop pairing (including with manual indexedDB clearing), vault operations, authentication requirement for desktop connection with browser restarts (memory clearing).

### dependencies

- https://github.com/tetherto/pearpass-app-desktop/pull/88


src/background/index.js
```diff
@@ -244,6 +228,31 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
...
+        if (!masterPassword) {
+          sendResponse({
+            success: false,
+            error: AUTH_ERROR_PATTERNS.MASTER_PASSWORD_REQUIRED,
+            code: ERROR_CODES.AUTHENTICATION_FAILED
+          })
+          return
+        }
+        await ensureClientKeypairUnlocked(masterPassword)
+        sendResponse({ success: true })
```
alvaropaco:
```suggestion
        return sendResponse({ success: true })
```

src/background/secureChannel.js
```diff
@@ -133,6 +132,25 @@ export class SecureChannelClient {
...
 
+    // Check if the client keystore is ready for signing. If not, we cannot
+    // complete a handshake that requires client signature. Throw early so
+    // callers know to wait for master password unlock.
+    try {
+      const { privateKey } = await ensureClientKeypairUnlocked()
+      if (!privateKey) {
+        throw new Error(AUTH_ERROR_PATTERNS.MASTER_PASSWORD_REQUIRED)
+      }
+    } catch (e) {
```
alvaropaco: Do we know that the e.message or e.stacktrace do not contains sensitive information?

src/background/secureChannel.js
```diff
@@ -361,8 +416,31 @@ export class SecureChannelClient {
...
+
+      // Reconstruct transcript used by desktop to derive session key
+      if (!this._session.hostEphemeralPubB64 || !this._ephemeralKeyPair) {
+        throw new Error(SESSION_ERROR_PATTERNS.HANDSHAKE_FAILED)
+      }
+
+      const hostEphemeralPublicKeyBytes = base64Decode(
+        this._session.hostEphemeralPubB64
+      )
+      const transcript = concatUint8Arrays([
```
alvaropaco: A suggestion here:
Prefix with a constant tag + version:
```javscript
const tag = new TextEncoder().encode('pearpass/handshake/v1/clientFinish')
const transcript = concatUint8Arrays([tag, hostEphemeralPublicKeyBytes, this._ephemeralKeyPair.publicKey])
```
This is a classic protocol-hardening move and costs almost nothing.

src/shared/utils/base64.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Encode a Uint8Array as standard Base64 (no URL-safe conversion).
+ * Works in extension service workers without Buffer.
+ * @param {Uint8Array} uint8Array
+ * @returns {string}
+ */
+export const base64Encode = (uint8Array) => {
+  const binary = String.fromCharCode.apply(null, uint8Array)
+  return btoa(binary)
```
alvaropaco: Wrap atob() with try/catch and throw a stable error.

src/background/index.js
```diff
@@ -244,6 +228,31 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
...
+      } catch (e) {
+        sendResponse({
+          success: false,
+          error: e?.message,
+          code: ERROR_CODES.AUTHENTICATION_FAILED
+        })
+      }
+    })()
+    return true
```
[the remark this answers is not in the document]
shavtvalishvili: Tells Chrome to keep the message channel open for async responses.

src/background/index.js
```diff
@@ -244,6 +228,31 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
     return true
   }
 
+  if (msg.type === SECURE_MESSAGE_TYPES.UNLOCK_CLIENT_KEYSTORE) {
```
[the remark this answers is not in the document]
shavtvalishvili: Already validated by validateSender

src/background/secureChannel.js
shavtvalishvili: IndexedDB errors may contain file paths, crypto and decode errors also may contain data details which is better to avoid, so I'm making these logs static.

PR #33, opened 2026-01-16
feat: http https optional restriction

harrisaar-noxtton:
## Feature

Adds configurable HTTP/HTTPS option in browser action popup settings, enabling users to select their preferred protocol at runtime.

## Architecture

A new boolean setting using `chrome.storage` browser API, consistent with other settings in the extension.

Popup context uses ActiveTab permission defined in `manifest.json` to verify page protocol matches security settings. Warning appears only when on HTTP with strict HTTPS required.

Content scripts use chrome.storage API with window.location (instead of active tab) to determine URL. A single function checks if the content script should be enabled. This function is called in multiple locations.

## Aspects to consider

1. 
In my opinion, we shouldn’t implement this feature. We can simply restrict manifest.json to enable the extension only on HTTPS websites. This would make the code more linear and simpler to maintain. I don’t believe many users actually need this feature—most would only use this browser manager on secure URLs.

2.
`activeTab` permission was used, because the browser action popup lacks access to the URL. Adding additional permissions might not be ideal approach. But it was a cleaner, simpler solution. Alternatives exist, like sending data from content scripts. 

3.
The settings are stored in `chrome.storage`, which is not synchronized between devices. If there's a future need to sync them, we'll need a new architecture. But this isn't directly related to this PR, as other settings use the same chrome.storage API approach.

4.
UX-wise, the settings are shown after the user pairs the extension and enters the master password. Under the hood, though, chrome.storage access does not need authentication. Users might assume these are tied to a specific  account, but they’re not.

5.
This setting allows using the browser extension on insecure domains like HTTP, but note that not all browser API features work there. For example, certain Chrome crypto API functions don’t on non-secure domains, and passkeys likely won’t either. It’s industry standard to limit security features to HTTPS, so some current and future features may not work.

## Video

Video demonstration of UX

[demonstrating http https setting .webm](https://github.com/user-attachments/assets/bd6145b2-48da-4910-8cd6-02edd2ec3da0)

### DIFF
--- a/src/shared/components/ButtonLittle/__snapshots__/index.test.jsx.snap
+++ b/src/shared/components/ButtonLittle/__snapshots__/index.test.jsx.snap
@@ -2,7 +2,7 @@
 
 exports[`ButtonLittle Component renders correctly when disabled 1`] = `
 <button
-  class="peer flex h-fit cursor-pointer items-center gap-2 text-sm font-medium focus:outline-none bg-primary400-mode1 px-2.5 py-1 rounded-[10px] text-black-mode1 hover:bg-primary500-mode1 pointer-events-none opacity-50"
+  class="peer flex h-fit cursor-pointer items-center gap-2 text-sm font-medium focus:outline-none bg-primary400-mode1 px-2.5 py-1 rounded-[10px] text-black-mode1 hover:bg-primary500-mode1 pointer-events-none opacity-50 "
   disabled=""
   type="button"
 >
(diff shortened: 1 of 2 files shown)


src/shared/hooks/useAllowHttpEnabled.js
torkinos: Also here 🙌

public/manifest.json
```diff
@@ -8,11 +8,12 @@
     "clipboardRead",
     "clipboardWrite",
     "offscreen",
-    "alarms"
+    "alarms",
+    "activeTab"
```
torkinos: We will probably need a description for this permission
@chetasr could you please check:

> activeTab
> This permission allows the extension to temporarily access the currently active browser tab when you interact with the extension. It is required to detect whether the website you are visiting uses a secure HTTPS connection or an unsecured HTTP connection for security awareness purposes. This access is limited to the active tab only and is not retained after you switch tabs or pages.

PR #35, opened 2026-01-19
Implemented Passkey feature

denislupookov:
### Requirements

- Implement passkey logic

### Changes

- Added creation/authentication logic for passkeys
- Added support for conditional UI to autofill passkey on request
- Fixed vulnerability where passkey private key was exposed to web pages during registration. Private key now stays within background/popup extension context and is sanitized via `sanitizeCredentialForPage()` before being sent to content scripts.
- Updated assertion signing logic to return DER-encoded ECDSA signatures per WebAuthn specification
- Added login screen if the vault is closed

### Testing Notes

Enable `PASSKEY_SUPPORT_ENABLED`

Good websites to test it:

https://demo.yubico.com/webauthn-developers (good for debugging)
https://www.passkeys-debugger.io/
https://passkeys.eu/

### Screenshots/Recordings

Basic example with creation/authentication using passkey

https://github.com/user-attachments/assets/18ce7264-cb2f-4a67-843e-f12f207d925f

Example with creation/authentication using autofill passkey

https://github.com/user-attachments/assets/1f8ddfdc-5e5b-4191-b364-b8d98517b88f

Example with overriding login on creation:

https://github.com/user-attachments/assets/fe127303-0f74-42e3-8080-a5a36548559a

### DIFF
--- a/src/action/containers/PasskeyContainer/index.jsx
+++ b/src/action/containers/PasskeyContainer/index.jsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState } from 'react'
+import { useEffect, useMemo, useRef, useState } from 'react'
 
 import { t } from '@lingui/core/macro'
 import {
@@ -17,6 +17,7 @@ import { ArrowDownIcon } from '../../../shared/icons/ArrowDownIcon'
 import { HardwareKey } from '../../../shared/icons/HardwareKey'
 import { LockCircleIcon } from '../../../shared/icons/LockCircleIcon'
 import { UserIcon } from '../../../shared/icons/UserIcon'
+import { isSameOrSubdomain } from '../../../shared/utils/isSameOrSubdomain'
 import { logger } from '../../../shared/utils/logger'
 import { normalizeUrl } from '../../../shared/utils/normalizeUrl'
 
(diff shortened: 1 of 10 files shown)


src/contentPopups/views/Autofill/index.jsx
```diff
@@ -57,67 +147,126 @@ export const Autofill = () => {
...
+
+    chrome.runtime.sendMessage(
+      {
+        type: MESSAGE_TYPES.AUTHENTICATE_WITH_PASSKEY,
+        credential: record.data.credential,
+        tabId: currentTabId
+      },
+      () => {
+        if (chrome.runtime.lastError) {
+          console.error('Failed to authenticate:', chrome.runtime.lastError)
```
torkinos: lets use `logger.error`
Also please check not to accidentally expose any sensitive stuff in error logs

src/action/containers/PasskeyContainer/index.jsx
```diff
@@ -0,0 +1,246 @@
...
+
+    const rpId = publicKeyData.rp?.id || publicKeyData.rpId
+    const normalizedRpId = rpId ? normalizeUrl(rpId, true) : null
+
+    let filtered = records.filter((record) => {
+      const websites = record.data?.websites || []
+      if (!normalizedRpId) return true
+
+      return websites.some((site) => {
+        const normalizedSite = normalizeUrl(site, true)
```
alvaropaco: This can cause false positives, e.g.:
- example.com matches notexample.com
- app.example.com matches example.com (which might be OK)
- evil-example.com matches example.com (not OK)
This is a common “substring domain matching” trap.

**Best practice**
Use proper host parsing and suffix rules:
- Convert both into a normalized hostname (not full URL).
- Match exact host OR valid subdomain relationship.

Pseudo:
```js
const isSameOrSubdomain = (a, b) => a === b || a.endsWith(`.${b}`)
```
Then:
```js
isSameOrSubdomain(normalizedSiteHost, normalizedRpIdHost) ||
isSameOrSubdomain(normalizedRpIdHost, normalizedSiteHost)
```
You can still allow root-domain records to match subdomains, but avoid substring matches.

PR #39, opened 2026-01-21
Add master password check for passkeys

denislupookov:
### Requirements

According to the WebAuthn specification, if UV is required, the authenticator must perform user verification or return an error. This verification must be added.

### Changes

- Added a settings option that lets the user decide when the master password should be requested.
- Added verification of master password in the popup

### Screenshots/Recordings

https://github.com/user-attachments/assets/3bbb62b4-0240-4295-bd44-c22d3c43d154

### DIFF
--- a/src/action/pages/CreatePasskey/index.jsx
+++ b/src/action/pages/CreatePasskey/index.jsx
@@ -29,13 +29,16 @@ export const CreatePasskey = () => {
         serializedPublicKey
       })
 
-      const { credential } = response
+      const { credential, publicKey } = response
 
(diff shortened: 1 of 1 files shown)


boris91 (approved):

src/shared/utils/passkeyVerificationPreference.js
torkinos: @denislupookov please also add tests

torkinos (changes requested):

torkinos (approved):

PR #75, opened 2026-02-06
Implement screen when there are no results after item searching

nikushaomanidze:
### Changes

When search yields no results, show only a "No result found." message instead of the empty collection view with create buttons.

### Testing Notes

Mac os chrome browser

### Screenshots/Recordings

Screenshot 2026-02-06 at 18 35 28

---
  - https://app.asana.com/0/0/1212855407829965


src/action/containers/EmptyCollectionView/index.jsx
```diff
@@ -5,7 +5,7 @@ import { ButtonCreate } from '../../../shared/components/ButtonCreate'
 import { RECORD_ICON_BY_TYPE } from '../../../shared/constants/recordIconByType'
 import { useRouter } from '../../../shared/context/RouterContext'
 
-export const EmptyCollectionView = () => {
+export const EmptyCollectionView = ({ isSearchActive }) => {
```
torkinos: add JS docs Pease @nikushaomanidze

PR #158, opened 2026-04-23
[Browser-Extension]:  Redesign Create new Vault flow (rename as well)

giorgikh93:
### Screenshots/Recordings

Screenshot 2026-04-23 at 15 05 31

Screenshot 2026-04-23 at 15 04 39

### dependencies

---
  - https://app.asana.com/0/0/1214113487981804

### DIFF
--- a/src/shared/containers/CreateOrEditVaultModalContentV2/index.tsx
+++ b/src/shared/containers/CreateOrEditVaultModalContentV2/index.tsx
@@ -1,0 +1,225 @@
+import type { ChangeEvent } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
+
+import { t } from '@lingui/core/macro'
+import { useForm } from '@tetherto/pear-apps-lib-ui-react-hooks'
+import { Validator } from '@tetherto/pear-apps-utils-validator'
+import {
+  useCreateVault,
+  useVault,
+  type Vault
+} from '@tetherto/pearpass-lib-vault'
+import {
+  AlertMessage,
+  Button,
+  DialogSurface,
+  Form,
+  InputField
+} from '@tetherto/pearpass-lib-ui-kit'
+
+import { useLoadingContext } from '../../context/LoadingContext'
+import { useRouter } from '../../context/RouterContext'
+import { logger } from '../../utils/logger'
+
+export type CreateOrEditVaultModalContentV2Props = {
+  onClose: () => void
+  onSuccess?: () => void
+  vault?: Vault
+}
+
+const nameSchema = Validator.object({
+  name: Validator.string().required(t`Name is required`)
+})
+
+export const CreateOrEditVaultModalContentV2 = ({
+  onClose,
+  onSuccess,
+  vault
+}: CreateOrEditVaultModalContentV2Props) => {
+  const { navigate } = useRouter()
+  const { isLoading, setIsLoading } = useLoadingContext()
+  const { createVault, isLoading: isCreateVaultLoading } = useCreateVault()
+  const { addDevice, updateUnprotectedVault } = useVault()
+
+  const isRename = Boolean(vault?.id)
+
+  const [submitError, setSubmitError] = useState<string | null>(null)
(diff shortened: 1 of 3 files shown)


boris91 (dismissed):

src/shared/containers/CreateOrEditVaultModalContentV2/index.tsx
```diff
@@ -0,0 +1,225 @@
...
+      )
+      setSubmitError(t`Could not create the vault. Try again.`)
+    }
+  }
+
+  const runSubmit = handleSubmit(submit)
+
+  return (
+    <div className="border-border-primary bg-surface-primary w-[500px] overflow-y-auto rounded-[8px] border">
+      <DialogSurface
```
denislupookov: but why DialogSurface not the Dialog component ?

giorgikh93: Dialog already includes a full-screen layer and its own backdrop. This screen is opened with our existing ModalContext, which already renders an overlay behind the content. Using Dialog would stack a second backdrop and duplicate the modal shell. DialogSurface is only the dialog card (header, body, footer), so we get the same UI kit styling without doubling the overlay or the outer wrapper.

denislupookov: okey, that sounds weird for me. Why should we keep the existing modal context ? Is there any logic we need that current one does not have ?

giorgikh93: I am not sure about logic. What was your purpose of creating 2 dialogs?

giorgikh93: Swapped DialogSurface with Dialog as the DialogSurface isn't intended to use as a modal - only in a rare specific cases. @denislupookov will provide some instruction or rename the component to avoid confusion in the future.

src/shared/containers/CreateOrEditVaultModalContentV2/index.tsx
```diff
@@ -0,0 +1,225 @@
...
+          <div className="flex w-full justify-end gap-[var(--spacing8)]">
+            <Button
+              variant="secondary"
+              size="small"
+              onClick={onClose}
+              data-testid="create-or-edit-vault-discard"
+            >
+              {t`Discard`}
+            </Button>
+            <Button
```
alvaropaco: This Button component has type as ["button" by default](https://github.com/tetherto/pearpass-lib-ui-react-native-components/blob/3dc00d071e575047dad5003d7351ed3d2318ea90/src/components/Button/Button.tsx#L59). I think it will not work as submit button. Maybe you need to set type as submit.

src/shared/containers/CreateOrEditVaultModalContentV2/index.tsx
```diff
@@ -0,0 +1,225 @@
...
+        error
+      )
+      setSubmitError(t`Could not create the vault. Try again.`)
+    }
+  }
+
+  const runSubmit = handleSubmit(submit)
+
+  return (
+    <div className="border-border-primary bg-surface-primary w-[500px] overflow-y-auto rounded-[8px] border">
```
denislupookov: I think we no longer need it

src/shared/containers/CreateOrEditVaultModalContentV2/index.tsx
denislupookov: why do we need submit type ? it is not a form

giorgikh93: @alvaropaco I think this recording answers your point, thanks.
 https://github.com/user-attachments/assets/fbec91f7-4f72-44a1-8b35-f91b3e06437f

alvaropaco (approved):

denislupookov (approved):

boris91 (approved):

PR #161, opened 2026-04-24
Redesign Move to Folder dialog (V2)

alvaropaco:
Introduces MoveFolderModalContentV2 behind isV2(), ported from the desktop V2 layout. Adds a chip-row destination picker (All Items / Favorites / custom folders) wired to updateRecords / updateFavoriteState / updateFolder. Keeps V1 path untouched.

### Changes

- designVersion.js with isV2() gated by EXTENSION_DESIGN_VERSION
- useRecordActionItems branches on isV2() and passes hasOverlay: false modal params so the kit Dialog's backdrop isn't stacked under the ModalContext overlay
- tetherto-modules.d.ts ambient declarations for untyped @tetherto/* packages (vault, avatar-initials, validator, ui-react-hooks); typed shape for useRecords, useFolders, useCreateFolder
- kit-theme.css imports kit design-system tokens (colors, spacing, radius, font-size) into the cascade so Tailwind utilities bg-surface-*, border-border-*, text-text-*, and CSS vars var(--spacing*) resolve at build time
- @testing-library/dom 10.4.1 devDep (peer of @testing-library/react)
- 11 component test cases covering chip ordering, disabled-at- destination, each submit path including Favorites option-c semantics (updateFavoriteState without touching folder), Discard, error alert; 2 new hook tests covering V1/V2 branch routing with modal params

### Screenshots/Recordings

https://github.com/user-attachments/assets/3d608960-92a6-482c-90ad-e55ce10e3a0e

### DIFF
--- a/src/shared/containers/MoveFolderModalContentV2/index.tsx
+++ b/src/shared/containers/MoveFolderModalContentV2/index.tsx
@@ -1,0 +1,301 @@
+import { useCallback, useMemo, useRef, useState } from 'react'
+
+import { t } from '@lingui/core/macro'
+import { generateAvatarInitials } from '@tetherto/pear-apps-utils-avatar-initials'
+import {
+  AlertMessage,
+  Button,
+  Dialog,
+  Text,
+  useTheme
+} from '@tetherto/pearpass-lib-ui-kit'
+import { Folder, Layers, StarBorder } from '@tetherto/pearpass-lib-ui-kit/icons'
+import { useFolders, useRecords } from '@tetherto/pearpass-lib-vault'
+
+import { RecordAvatar } from '../../components/RecordAvatar'
+import { RECORD_COLOR_BY_TYPE } from '../../constants/recordColorByType'
+import { useLoadingContext } from '../../context/LoadingContext'
+import { useModal } from '../../context/ModalContext'
+import { logger } from '../../utils/logger'
+
+const CHIP_ID_ALL = '__all__'
+const CHIP_ID_FAVORITES = '__favorites__'
+
+export type MoveFolderRecord = {
+  id: string
+  type?: string
+  folder?: string | null
+  isFavorite?: boolean
+  data?: {
+    title?: string
+    username?: string
+    email?: string
+    websites?: string[]
+    [key: string]: unknown
+  }
+}
+
+export type MoveFolderModalContentV2Props = {
+  records: MoveFolderRecord[]
+  onCompleted?: () => void
+}
+
+type IconComponent = React.ComponentType<{ color?: string }>
+
+type ChipOption =
+  | { kind: 'all'; id: typeof CHIP_ID_ALL; label: string; Icon: IconComponent }
+  | {
+      kind: 'favorites'
+      id: typeof CHIP_ID_FAVORITES
+      label: string
+      Icon: IconComponent
+    }
+  | { kind: 'custom'; id: string; label: string; Icon: IconComponent }
+
+function getRecordSubtitle(record: MoveFolderRecord): string {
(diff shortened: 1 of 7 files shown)


boris91 (dismissed):

src/shared/containers/MoveFolderModalContentV2/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+            size="small"
+            title={t`Something went wrong`}
+            description={submitError}
+            testID="move-folder-v2-alert"
+          />
+        ) : null}
+
+        {previewRecord ? (
+          <div
+            className="flex items-start gap-[var(--spacing12)] rounded-[var(--radius8)] p-[var(--spacing12)]"
```
denislupookov, line 230: can you use ListItem here

src/shared/containers/MoveFolderModalContentV2/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+                {previewRecord.data?.title ?? ''}
+              </Text>
+              <Text variant="body" color={theme.colors.colorTextSecondary}>
+                {getRecordSubtitle(previewRecord)}
+              </Text>
+            </div>
+          </div>
+        ) : null}
+
+        <div
```
denislupookov, line 254: why we do not use Button here ?

src/shared/containers/MoveFolderModalContentV2/index.tsx
alvaropaco: Yes we could, but will not be piexel by piexel aligned with the FIgma.
The preview row is display only. It shows the record being moved, not something the user selects. Kit ListItem wraps Pressable internally, so even with selectable={false} it renders as an interactive node and ATs may announce it as such. It also applies its own typography and container chrome, so matching the Figma padding/text variants would mean threading titleStyle/subtitleStyle overrides. The custom div + RecordAvatar + Text composition mirrors what the desktop V2 uses for the same screen, keeping it aligned.

src/shared/containers/MoveFolderModalContentV2/index.tsx
alvaropaco: The chips are a radio group (mutually exclusive selection), not independent push-buttons; role="radio" is the correct semantic. Kit Button has no "selected" variant (only primary | secondary | tertiary | destructive), and its chrome doesn't match the Figma chip (padding: 12px, border-secondary on both states, surface-elevated-on-interaction bg when selected). Using Button would require fighting its styles and lose the radio-group a11y. Desktop's MoveFolderModalContentV2 uses the same <div role="radio"> pattern for the same reason.

src/shared/containers/MoveFolderModalContentV2/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+import { useModal } from '../../context/ModalContext'
+import { logger } from '../../utils/logger'
+
+const CHIP_ID_ALL = '__all__'
+const CHIP_ID_FAVORITES = '__favorites__'
+
+// Modal params passed to ModalContext.setModal so the kit Dialog's own
+// backdrop/focus-trap is the only one on screen (otherwise ModalContext
+// would stack its <Overlay/> on top of the Dialog's backdrop).
+export const MOVE_FOLDER_MODAL_V2_PARAMS = { hasOverlay: false } as const
```
denislupookov: let's get rid of it and inside the Modal Context remove overlay for v2 design please

denislupookov (approved):

boris91 (approved):

giorgikh93 (approved):

PR #165, opened 2026-04-27
[Browser-Extension] Redesign password field behaviour

giorgikh93:
### V2

https://github.com/user-attachments/assets/e9c9e751-6ea7-4a82-8984-4dd8ebe12799

### V1

https://github.com/user-attachments/assets/f32a7c2e-b832-4fd3-8590-93df88cc7688

### dependencies

---
  - https://app.asana.com/0/0/1214113487981834


boris91 (dismissed):

hakaari:
@nickgr2 please review the above

nickgr2 (changes requested):
A few things worth addressing before this lands.

### 1. `react-dom/server` shipped to the content script (`src/content/utils/showPasswordStrengthNearField.js`)
This is the first import of `react-dom/server` in `src/content/`. The renderer (and React itself) gets pulled into a bundle that runs on every navigated page. For four small icons this is a heavy cost. Two options:

- Inline the four SVGs as static strings (the kit icons only need `width`/`height`/`color`).
- Pre-render the icons once at build time and import the resulting strings.

Either way avoids dragging the React server renderer onto every page.

### 2. Hardcoded user-facing English strings (`getDisplayLabel`)
```js
if (type === PASSWORD_STRENGTH.SAFE) return 'Strong'
if (type === PASSWORD_STRENGTH.WEAK) return 'Decent'
return 'Vulnerable'
```
These render into the overlay and are user-visible. Per the i18n rule, route through Lingui. The content script doesn't use Lingui today, but `i18n._(t\`Strong\`)` etc. with the shared `i18n` instance keeps these translatable rather than baking English into a content-script bundle. If wiring Lingui into `src/content/` is out of scope for this PR, please open a follow-up issue and reference it from a TODO in the file (or, preferably, do it here).

### 3. Hardcoded hex colors in the overlay
`#C0D836`, `#15180E`, `#212814`, `#BEE35A`, `#D7D245`, `#141414`, `#1c1c1c`, `#D13B3D` are all literals in `showPasswordStrengthNearField.js`. Content scripts can't pull from the kit's CSS theme tokens at runtime, so a literal source is unavoidable — but please pull these into a single `STRENGTH_COLORS` constant (or a thin helper that mirrors `rawTokens` shape) so a future palette refresh is one edit, not a hunt across the file. Same idea for `font-family: Inter, system-ui, ...` and the `12px` font size.

### 4. Unnecessary type assertion in `PasswordGeneratorV2`
```ts
const indicatorVariant: PasswordIndicatorVariant =
  STRENGTH_TO_INDICATOR[(strength as { type: string }).type] ?? 'vulnerable'
```
The new `tetherto-modules.d.ts` declares `checkPasswordStrength` / `checkPassphraseStrength` returning `{ type: string; [key: string]: unknown }`. The `as` is dead weight — drop it: `STRENGTH_TO_INDICATOR[strength.type]`.

### 5. `PasswordSuggestionV2` refresh control isn't keyboard-accessible
```jsx
<div
  className="flex shrink-0 cursor-pointer items-center gap-[var(--spacing12)]"
  data-testid="passwordsuggestionv2-refresh"
  onClick={(e) => { e.stopPropagation(); onGeneratePassword() }}
>
  <SyncLock ... />
</div>
```
No `role`, no `tabIndex`, no `onKeyDown` — keyboard users can't trigger refresh. The fill row above does this correctly. Use the kit `Button` (icon variant) or mirror the role/keyboard pattern.

### 6. minor: `useEffect` deps in both V2 popups
Both `PasswordGeneratorV2` and `PasswordSuggestionV2` declare `[recordsData?.length]` / `[filteredRecords?.length]` while the body reads `routerState`, `refetchVault`, `isLoading`, `isInitialized`, `popupRef.current`. v1 has the same shape, so this isn't new — but if the v2 work is the chance to clean it up, exhaustive deps would catch real stale-closure bugs later.

### 7. minor: tests
`showPasswordStrengthNearField.test.js` is solid. `PasswordGeneratorV2` and `PasswordSuggestionV2` have no tests — v1 didn't either, so this matches existing precedent, but adding even one render-and-click test per popup would lock in the strength-indicator wiring and the postMessage payload contract.

### Padding-restoration ordering — looks correct
For the record: `existing.cleanup()` runs before reading `inputElement.style.paddingRight`, so the new pill captures the original (pre-overlay) inline value. No bug there.

giorgikh93:
@nickgr2 I've addressed your suggestions except point 5 and 6,  please review again. 
I didn't address `point 5` because: we don't use the pure html buttons for the consitency sake, instead the one from the @tetherto/pearpass-lib-ui-kit which currently lacks borderdless option and I am bit hurry to merge this PR.

`point 6` because: refetchVault from useVault() isn’t referentially stable, it’s often a new function each render. If we list it in the effect deps, the effect re-runs every render, calls refetch again, state updates, re-render, new refetch - loop. We only need to re-run when something meaningful changes (e.g. recordsData?.length), so we don’t add refetchVault to the deps unless the library guarantees a stable refetch (e.g. useCallback in the provider).

nickgr2:
> I didn't address `point 5` because: we don't use the pure html buttons ... currently lacks borderdless option and I am bit hurry to merge this PR.

The kit `Button` isn't the only path — the original review listed "or mirror the role/keyboard pattern". Three lines on the existing `div` covers it: `role="button"`, `tabIndex={0}`, and an `onKeyDown` that fires `onGeneratePassword()` on Enter/Space. No kit migration, no borderless dependency. Please do it before merge — the refresh control is otherwise unreachable for keyboard-only users.

> refetchVault from useVault() isn't referentially stable

Fair, accepted. Standard escape hatch (`useRef` mirroring the latest function) would let it satisfy exhaustive-deps without re-running, but that's a follow-up at best — not blocking here.

nickgr2 (changes requested):
Heavy lifting from the prior review is in — `react-dom/server` is gone (inline SVG strings), `t``...``` is now wired through `contentI18n.js`, and the hex literals collapsed into `STRENGTH_COLORS` / `STRENGTH_TYPOGRAPHY` / `THEME`. Test coverage for both V2 popups was added too. Nice response.

Two items from the prior review are still open:

### `PasswordSuggestionV2` refresh control still isn't keyboard-accessible
`src/contentPopups/views/PasswordSuggestionV2/index.tsx:120-130`:

```jsx
<div
  className="flex shrink-0 cursor-pointer items-center gap-[var(--spacing12)]"
  data-testid="passwordsuggestionv2-refresh"
  onClick={(e) => {
    e.stopPropagation()
    onGeneratePassword()
  }}
>
  <SyncLock ... />
</div>
```

No `role`, no `tabIndex`, no `onKeyDown` — keyboard-only and screen-reader users still can't reach refresh. The fill row above already shows the right pattern (`role="button"` + `tabIndex={0}` + `onKeyDown` handling Enter/Space). Mirror that, or use the kit `Button` (icon variant).

### Dead type assertion in `PasswordGeneratorV2`
`src/contentPopups/views/PasswordGeneratorV2/index.tsx:138-139`:

```ts
const indicatorVariant: PasswordIndicatorVariant =
  STRENGTH_TO_INDICATOR[(strength as { type: string }).type] ?? 'vulnerable'
```

`tetherto-modules.d.ts` declares `checkPasswordStrength` / `checkPassphraseStrength` returning `{ type: string; [key: string]: unknown }`, so `strength.type` is already `string`. The cast is dead weight — `STRENGTH_TO_INDICATOR[strength.type]` is enough.

---

Both small. Happy to approve once they're in.

giorgikh93:
@nickgr2 addressed your requests, please review!

nickgr2 (approved):
Both items from the previous review are in:

- `PasswordSuggestionV2` refresh now has `role="button"`, `tabIndex={0}`, an `aria-label` via `useLingui`/`t``...```, and an `onKeyDown` handler for Enter/Space mirroring the fill-row pattern. Wrapping the icon in `<span aria-hidden>` is a nice touch, and the test mock for `@lingui/react` keeps the suite green.
- `PasswordGeneratorV2` no longer carries the dead `(strength as { type: string }).type` cast — `STRENGTH_TO_INDICATOR[strength.type]` is enough given the module typing.

LGTM.

boris91 (approved):

torkinos (approved):

PR #168, opened 2026-04-27
Redesign folder flow

ndolidzee:
### Screenshots/Recordings

https://github.com/user-attachments/assets/1f867a7c-c122-42a4-8b25-c5d538884552

https://github.com/user-attachments/assets/c6bc8af0-040e-4bc1-8ac3-d7d2df78bc59

### DIFF
--- a/src/shared/containers/CreateFolderModalContentV2/index.tsx
+++ b/src/shared/containers/CreateFolderModalContentV2/index.tsx
@@ -1,0 +1,149 @@
+import type { ChangeEvent } from 'react'
+import { useState } from 'react'
+
+import { t } from '@lingui/core/macro'
+import { useForm } from '@tetherto/pear-apps-lib-ui-react-hooks'
+import { Validator } from '@tetherto/pear-apps-utils-validator'
+import { Button, Dialog, Form, InputField } from '@tetherto/pearpass-lib-ui-kit'
+import { useCreateFolder, useFolders } from '@tetherto/pearpass-lib-vault'
+
+import { useLoadingContext } from '../../context/LoadingContext'
+import { logger } from '../../utils/logger'
+
+export type CreateFolderModalContentV2Props = {
+  onClose: () => void
+  onCreate?: (folderName: string) => void
+  onRename?: (newFolderName: string, previousFolderName: string) => void
+  initialValues?: { title: string }
+}
+
+export const CreateFolderModalContentV2 = ({
+  onClose,
+  onCreate,
+  onRename,
+  initialValues
+}: CreateFolderModalContentV2Props) => {
+  const { setIsLoading } = useLoadingContext()
+  const isRename = !!initialValues
+
+  const { renameFolder, data } = useFolders()
+  const customFolders: Array<{ name: string }> = Object.values(
+    data?.customFolders ?? {}
+  )
+
+  const [isRenameLoading, setIsRenameLoading] = useState(false)
+
+  const { isLoading: isCreateLoading, createFolder } = useCreateFolder({
+    onCompleted: (folderName: string) => {
+      onCreate?.(folderName)
+      onClose()
+    }
+  })
+
+  const isLoading = isRename ? isRenameLoading : isCreateLoading
+
+  const schema = Validator.object({
+    title: Validator.string()
+      .required(t`Title is required`)
+      .refine((value: string): string | null => {
+        if (isRename && value === initialValues?.title) {
+          return null
+        }
+        const isDuplicate = customFolders.some(
+          (folder) => folder.name === value
+        )
+        if (isDuplicate) {
+          return t`Folder already exists`
+        }
+        return null
+      })
+  })
+
+  const { register, handleSubmit, setValue } = useForm({
+    initialValues: {
+      title: initialValues?.title ?? ''
+    },
+    validate: (formValues: Record<string, string>) =>
+      schema.validate(formValues) as Record<string, string>
+  })
+
+  const titleField = register('title')
+
+  const onSubmit = async (formValues: Record<string, string>) => {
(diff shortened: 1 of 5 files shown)


ndolidzee:
@nickgr2 please review the above

nickgr2 (changes requested):
A few correctness and convention issues to address.

### `DeleteFolderModalContentV2` should not have the `V2` suffix
[`src/shared/containers/DeleteFolderModalContentV2/index.tsx`](src/shared/containers/DeleteFolderModalContentV2/index.tsx) — there is no v1 sibling `DeleteFolderModalContent` in `src/shared/containers/`. Per `AGENTS.md` ("File naming: when to use the `V2` suffix"):

> No v1 equivalent exists (net-new feature, net-new component) → create the file with its natural name, **no `V2` suffix**. The suffix would just be noise.

Rename the directory and component to `DeleteFolderModalContent` and update the testIDs (`deletefolder-*-v2` → drop the `-v2`) along with the import in `Sidebar/index.jsx`. The rename modal (`CreateFolderModalContentV2`) is fine as-is since v1 `CreateFolderModalContent` exists.

### Validation runs against the untrimmed value, but the trimmed value is persisted
[`CreateFolderModalContentV2/index.tsx`](src/shared/containers/CreateFolderModalContentV2/index.tsx) lines 51–66 and 80: the duplicate-name `refine` runs on the raw form value, but `onSubmit` calls `trimmed` into `createFolder` / `renameFolder`. So a user entering `"Work "` (trailing space) when a folder `"Work"` already exists will pass validation, then attempt to create/rename to `"Work"` — collision at the storage layer. Trim before checking:

```ts
.refine((value: string): string | null => {
  const candidate = value.trim()
  if (isRename && candidate === initialValues?.title) return null
  if (customFolders.some((folder) => folder.name === candidate)) {
    return t`Folder already exists`
  }
  return null
})
```

Same applies to the empty-name guard (`isSaveDisabled`) — already uses trimmed value, so it's fine.

### `declare module '@tetherto/pearpass-lib-constants'` with no body may shadow real types
[`src/tetherto-modules.d.ts`](src/tetherto-modules.d.ts) line 10 — adding an empty ambient declaration like this turns every export from the module into `any`. This isn't just `UNSUPPORTED` in the new file — it also affects existing typed imports from `src/action/pages/SettingsV2/content/AppPreferencesContent/index.tsx` and any future `.tsx` consumer.

If the package ships its own `.d.ts` files, drop this line entirely; if it doesn't, declare the specific exports you need (mirroring the pattern used for `@tetherto/pear-apps-utils-avatar-initials` further down in the same file). Quick fix:

```ts
declare module '@tetherto/pearpass-lib-constants' {
  export const UNSUPPORTED: boolean
  // ...other exports as needed
}
```

### `UNSUPPORTED` is an opaque flag name
`DeleteFolderModalContentV2` lines 4, 75, 92 — naming-wise, `!UNSUPPORTED` reads as "if not unsupported" with no clue about *what* is unsupported. Other flags in this package (`AUTHENTICATOR_ENABLED`, `PROTECTED_VAULT_ENABLED`, `SAVE_CREDENTIALS_AFTER_LOGIN_ENABLED`) are feature-specific. If this constant lives upstream in `pearpass-lib-constants`, please rename it there to something like `DELETE_FOLDER_PRESERVE_ITEMS_ENABLED` (and flip the polarity) before this lands — otherwise the next reader has to grep the constant to find what's gated.

### Empty-folder deletion skips confirmation
[`Sidebar/index.jsx`](src/shared/containers/Sidebar/index.jsx) lines 154–160 — when an empty folder is deleted, the action fires immediately with no dialog. v1 always asked for confirmation. If the redesign intends "no items, no risk → no prompt", that's a fine call but please confirm; otherwise route empty folders through the same modal (or a smaller "Delete folder?" confirmation) so the click isn't a single-step destructive action.

### `minor:` Inconsistent `state` access in Sidebar
[`Sidebar/index.jsx`](src/shared/containers/Sidebar/index.jsx) line 156 uses `state?.folder` while line 131 uses `state.folder` directly. Same hook, same scope — pick one style. If `state` can be undefined here, line 131 is also unsafe; if not, drop the `?.`.

### `minor:` Verbose router type cast in `DeleteFolderModalContentV2`
Lines 199–209 — the inline `as { params: ...; state: ... }` casts to `navigate` and `(state as { folder?: string } | undefined)?.folder` are awkward. Since `RouterContext` is plain JSX and untyped, prefer a tiny local type alias at the top of the file (or thread the folder name through a typed selector) rather than per-call casts.

### Notes on what's right
- testIDs on every interactive element ✅
- All strings via Lingui `t\`\`` ✅
- Kit-only imports, no new files under `src/shared/components/` ✅
- Tailwind tokens via `var(--spacing8)`, no raw px/hex ✅

ndolidzee:
@nickgr2 I've addressed your suggestions. Please review again
UNSUPPORTED flag is currently known for the whole team and there is no issue using it we can skip that.
Also skipping confirmation on folder delete is intentional.

nickgr2 (approved):
All prior changes addressed — looks good.

- `DeleteFolderModalContentV2` → `DeleteFolderModalContent` (directory, type, export, testIDs)
- Trim-before-check in `CreateFolderModalContentV2` duplicate-name validation
- `@tetherto/pearpass-lib-constants` ambient declaration now lists explicit exports
- Sidebar `state.folder` access consistent
- Inline router casts replaced with `RouterState` / `RouterNavigateArg` aliases applied once at `useRouter()`

Per author confirmation, leaving `UNSUPPORTED` naming and the empty-folder no-confirm path as intentional.

boris91 (approved):

torkinos (approved):

PR #170, opened 2026-04-27
feat: App Version and Language pages

alvaropaco:
Rebuilt two settings screens — Language and App Version — to match the new design.

https://github.com/user-attachments/assets/977b80df-4bfa-4b81-872d-639daff31db0

Screenshot 2026-04-27 at 16 04 44
Screenshot 2026-04-27 at 16 04 36

---
  - https://app.asana.com/0/0/1214113487981828

### DIFF
--- a/src/action/index.jsx
+++ b/src/action/index.jsx
@@ -14,12 +14,21 @@ import { LoadingProvider } from '../shared/context/LoadingContext'
 import { ModalProvider } from '../shared/context/ModalContext'
 import { RouterProvider } from '../shared/context/RouterContext'
 import { ToastProvider } from '../shared/context/ToastContext'
+import { getLocaleFromStorage } from '../shared/utils/localeStorage'
 import { logger } from '../shared/utils/logger'
 import '../index.css'
 import '../strict.css'
 
 i18n.load('en', messages)
 i18n.activate('en')
+
+getLocaleFromStorage()
+  .then((stored) => {
+    if (stored && stored !== i18n.locale) i18n.activate(stored)
+  })
+  .catch((error) => {
+    logger.error('Failed to load persisted locale:', error)
+  })
 
 createClient()
   .then(() => {
--- a/src/action/pages/SettingsV2/content/AppVersionContent/index.tsx
+++ b/src/action/pages/SettingsV2/content/AppVersionContent/index.tsx
@@ -66,7 +66,7 @@ export const AppVersionContent = () => {
 
       <div
         data-testid={TEST_IDS.versionRow}
-        className="border-border-primary flex w-full items-center justify-between rounded-[8px] border px-[12px] py-[15px]"
+        className="border-border-primary flex w-full items-center justify-between rounded-[8px] border px-[12px] py-[16px]"
       >
         <Text variant="labelEmphasized">{t`App version`}</Text>
         <Text
--- a/src/action/pages/SettingsV2/content/AppVersionContent/index.tsx
+++ b/src/action/pages/SettingsV2/content/AppVersionContent/index.tsx
@@ -1,11 +1,82 @@
 import { t } from '@lingui/core/macro'
-import { PageHeader } from '@tetherto/pearpass-lib-ui-kit'
+import { Trans } from '@lingui/react/macro'
+import {
+  PEARPASS_WEBSITE,
+  PRIVACY_POLICY,
+  TERMS_OF_USE
+} from '@tetherto/pearpass-lib-constants'
+import { Link, PageHeader, Text, useTheme } from '@tetherto/pearpass-lib-ui-kit'
 
-export const AppVersionContent = () => (
-  <PageHeader
-    as="h1"
-    testID="settings-app-version"
-    title={t`App version`}
-    subtitle={t`Here you can find all the info about your app.`}
-  />
-)
+import { version } from '../../../../../../public/manifest.json'
+
+const TEST_IDS = {
+  root: 'settings-app-version',
+  termsLink: 'settings-app-version-terms-link',
+  privacyLink: 'settings-app-version-privacy-link',
+  websiteLink: 'settings-app-version-website-link',
+  versionRow: 'settings-app-version-row',
+  versionValue: 'settings-app-version-value'
+} as const
+
+export const AppVersionContent = () => {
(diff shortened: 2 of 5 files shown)


boris91 (dismissed):

giorgikh93:
@nickgr2 please review this PR

nickgr2 (commented):
Both screens follow the kit-first pattern established in `AppPreferencesContent` — `Dropdown` + `Button` + `NavbarListItem` composition, `useTheme()` for token colors, Tailwind utilities (`bg-surface-primary`, `border-border-primary`), `KeyboardArrowBottom` from the kit's icon subpath, no v1 imports, full Lingui coverage with `<Trans>` for the link-bearing subtitle, `data-testid` on every interactive element. Net-new files use natural names (no `V2` suffix) — correct given there's no v1 sibling in `SettingsV2/content/`.

A few items for the author to confirm or address:

1. **Locale persistence (carry-over from v1, but worth fixing here).** `LanguageContent` calls `i18n.activate(option.value)` and updates local state, but the choice isn't written anywhere persistent. `src/action/index.jsx` hardcodes `i18n.activate('en')` on every popup mount, so a user's language selection is lost the next time the popup opens. The v1 `AppearanceContent` had the same gap, so this isn't a regression — but since the redesign is the right moment to land the fix, please consider persisting via `chrome.storage.local` (or `localStorage`) and reading it from the popup entry point. If that's out of scope for this PR, file a follow-up so it isn't forgotten.

2. **`py-[15px]` in AppVersionContent (`index.tsx:69`).** Vertical padding on the version row is `15px`, but the kit's spacing scale jumps `12 → 16 → 20`. AGENTS.md notes feature-specific layout literals are OK, but `15px` reads like it might want to be `spacing16` (`py-[var(--spacing16)]` / `py-[16px]`). Confirm 15px is what the design specifies; if not, snap to the nearest token.

3. **`useLanguageOptions` cast (`LanguageContent/index.tsx:30-32`).** `as { languageOptions: LanguageOption[] }` mirrors the pattern in `AppPreferencesContent` for casting JS-hook returns, so this is consistent with the codebase. Longer-term it'd be nicer to add JSDoc types to `useLanguageOptions.js` so consumers don't need the cast — not blocking here.

4. **Tests.** Neither component ships with a test file. `AppPreferencesContent` already has a test that mocks the `Dropdown` / `KeyboardArrowBottom` flow — `LanguageContent` could reuse that scaffold to cover the "select option → handleSelect fires → `i18n.activate` is called" path, and `AppVersionContent` deserves at least a smoke test asserting the version row renders `{version}` and the three external links carry `isExternal`. `ReportAProblemContent` and `YourVaultsContent` also lack tests, so this is a wish rather than a block.

5. **Minor — dropdown chevron state.** `KeyboardArrowBottom` doesn't rotate when the dropdown opens. Consistent with `AppPreferencesContent`, so no change requested — flagging only in case the design specifies a flipped state for this screen.

Approving — items above are improvements, not blockers.

alvaropaco:
@nickgr2 please review this PR again after the concern addressed:
In-scope changes (5 files):

[src/shared/constants/storage.js](vscode-webview://0pb5ovl2ua2akv0mebbsm43h24toinf485c21nlbhogvdkmsdpk8/pearpass-app-browser-extension/src/shared/constants/storage.js) — added LOCALE: 'locale' to CHROME_STORAGE_KEYS
[src/shared/utils/localeStorage.js](vscode-webview://0pb5ovl2ua2akv0mebbsm43h24toinf485c21nlbhogvdkmsdpk8/pearpass-app-browser-extension/src/shared/utils/localeStorage.js) — new file, mirrors allowHttpStorage.js pattern (getLocaleFromStorage / setLocaleInStorage)
[src/action/pages/SettingsV2/content/LanguageContent/index.tsx](vscode-webview://0pb5ovl2ua2akv0mebbsm43h24toinf485c21nlbhogvdkmsdpk8/pearpass-app-browser-extension/src/action/pages/SettingsV2/content/LanguageContent/index.tsx) — handleSelect now persists locale via setLocaleInStorage
[src/action/index.jsx](vscode-webview://0pb5ovl2ua2akv0mebbsm43h24toinf485c21nlbhogvdkmsdpk8/pearpass-app-browser-extension/src/action/index.jsx) — popup mount reads stored locale, re-activates if different from default 'en'. Errors logged via logger.error, no fatal block
[src/action/pages/SettingsV2/content/AppVersionContent/index.tsx:69](vscode-webview://0pb5ovl2ua2akv0mebbsm43h24toinf485c21nlbhogvdkmsdpk8/pearpass-app-browser-extension/src/action/pages/SettingsV2/content/AppVersionContent/index.tsx#L69) — py-[15px] → py-[16px] snapped to spacing16 token
Reviewer items handled:

Locale persistence ✅ (chrome.storage.local, also fixes v1 carry-over since both flows share same storage key going forward)
py-15 snap ✅ → 16
useLanguageOptions cast — left as-is (non-blocking, consistent with AppPreferencesContent)
Tests — out of scope per reviewer ("wish rather than block") + user constraint
Chevron rotate — non-blocking flag, left as-is
Verification: TypeScript clean for changed files (only pre-existing TS2786 noise project-wide); ESLint clean.

nickgr2 (approved):
The two items I flagged are addressed:

1. **Locale persistence** — new `src/shared/utils/localeStorage.js` exposes `getLocaleFromStorage` / `setLocaleInStorage` against `chrome.storage.local` (keyed off `CHROME_STORAGE_KEYS.LOCALE`). `LanguageContent` writes the choice in `handleSelect` after `i18n.activate`, and `src/action/index.jsx` re-reads on popup mount and re-activates the stored locale. The `chrome?.storage?.local?.get` / `set` guards are appropriate for the extension context. Note that `i18n.load('en', messages)` still only seeds English at module init, so picking it/es/fr will activate the locale without messages loaded — that's the same pre-existing limitation the language picker had before this PR, not something new.
2. **`py-[15px]` → `py-[16px]`** in `AppVersionContent` — done, snapped to the spacing scale.

Items 3–5 (cast cleanup, tests, chevron rotate) were non-blocking and unchanged here, which is fine.

Approving.

denislupookov (approved):

torkinos (approved):

boris91 (approved):

PR #171, opened 2026-04-27
feat: Redesign the error modals

alvaropaco:


alvaropaco:
@nickgr2 please review

nickgr2 (changes requested):
A few concerns, the Step 2 flow being the main one.

**Step 2 sync-failure UX is a dead end** (`src/onboarding/Step2Dialog.tsx:53-55`)

```tsx
if (hasSyncFailed) {
  return <SyncingFailedModal onRetry={handleConnect} />
}
```

When `getIdentity()` fails, the entire Step 2 view is replaced by `SyncingFailedModal`, which exposes only **Sync Again** — no close, no back, no way to edit the pair code. `onRetry` calls `handleConnect` again with the same `code` from state, so if the failure is a bad pair code (typo, expired, already-consumed) the user is stuck retrying the same bad code forever and can only escape by closing the popup window.

The previous flow surfaced the error inline via `AlertMessage` so the user could correct the code and retry in place. Suggest either (a) keeping the input field visible with the failure message rendered alongside it, or (b) giving `SyncingFailedModal` a secondary action (e.g. "Re-enter code") that flips `hasSyncFailed` back to `false` and returns to the Step 2 form.

**Loss of error context** (`src/onboarding/Step2Dialog.tsx:42-49`)

The previous catch block extracted the actual error message from desktop (`Failed to get identity. Please try again.` for `success: false`, or the thrown `err.message` for exceptions). The new code collapses every failure to "Ensure the PearPass desktop app is open and Browser Sync is enabled.", which is misleading when the underlying error is "code expired" or "code already used". If the design team wants a single failure surface, consider passing the upstream error through to `SyncingFailedModal` so the body text (or a sub-line) reflects the real cause.

**`SyncingFailedModal` doesn't follow the spacing-token convention** (`src/onboarding/SyncingFailedModal.tsx:35,37`)

```
className="bg-surface-primary flex w-full max-w-[500px] flex-col gap-[24px] rounded-[8px] p-[24px]"
...
className="flex flex-col items-center gap-[8px] text-center"
```

Every other file in `src/onboarding/` (and `PairingRequiredModalContentV2.tsx`, the closest sibling V2 modal) uses `gap-[var(--spacing24)]` / `gap-[var(--spacing8)]` etc., which map to the same pixel values. Worth aligning so this doesn't drift the convention. The new V2 modals (`NonSecureWarningV2.tsx:38`, `DesktopConnectionModalContentV2.tsx:69`) have the same `gap-[8px]` issue.

**Missing tests for the three new V2 components**

`PairingRequiredModalContentV2` ships with a colocated `*.test.tsx` and snapshots. None of `NonSecureWarningV2`, `DesktopConnectionModalContentV2`, or `SyncingFailedModal` have tests in this PR. The Step 2 flow change in particular (`hasSyncFailed` toggle, retry call shape) is the kind of behavior worth pinning with a test before merge.

**`DesktopConnectionModalContentV2.handleClose` always calls `window.close()`** (`...DesktopConnectionModalContentV2.tsx:34-41`)

Worth confirming this is intentional: the V2 dialog is registered with `closeable: true` and `Dialog` defaults to `closeOnOutsideClick: true`, so any of (Esc / backdrop click / X button / Discard) now closes the entire action popup. The V1 version only closed the modal. That seems to be the intent of the redesign but it's a behavior change worth a sanity check.

**minor:** PR title typo — "Redesing" → "Redesign".

alvaropaco:
@nickgr2 please review again please. I've fixed your reqested changes and raised points:
Concerns fixed:

Step 2 dead-end ([SyncingFailedModal.tsx](vscode-webview://0pb5ovl2ua2akv0mebbsm43h24toinf485c21nlbhogvdkmsdpk8/pearpass-app-browser-extension/src/onboarding/SyncingFailedModal.tsx) + [Step2Dialog.tsx](vscode-webview://0pb5ovl2ua2akv0mebbsm43h24toinf485c21nlbhogvdkmsdpk8/pearpass-app-browser-extension/src/onboarding/Step2Dialog.tsx)): added onCancel prop → renders "Re-enter code" secondary button. Step 2 passes () => setSyncErrorMessage(null) → flips state back to form, user edits code + retries.

Lost error context: Step2Dialog now tracks syncErrorMessage: string | null (replaces boolean). Extracts message from res.error / Error.message / fallback. Passes errorMessage prop to SyncingFailedModal → renders sub-line with caption color.

Spacing tokens: all hardcoded gap-[24px] / gap-[8px] / p-[24px] / rounded-[8px] → gap-[var(--spacing24)] / gap-[var(--spacing8)] / p-[var(--spacing24)] / rounded-[var(--radius8)]. Affects SyncingFailedModal.tsx, NonSecureWarningV2.tsx, DesktopConnectionModalContentV2.tsx.

Tests added (3 new files, 20 tests, 3 snapshots, all pass):

[SyncingFailedModal.test.tsx](vscode-webview://0pb5ovl2ua2akv0mebbsm43h24toinf485c21nlbhogvdkmsdpk8/pearpass-app-browser-extension/src/onboarding/SyncingFailedModal.test.tsx) — body copy, cancel visibility/click, error detail visibility, retry call
[NonSecureWarningV2.test.tsx](vscode-webview://0pb5ovl2ua2akv0mebbsm43h24toinf485c21nlbhogvdkmsdpk8/pearpass-app-browser-extension/src/action/pages/NonSecureWarning/NonSecureWarningV2.test.tsx) — body copy, closeOnOutsideClick=false, Discard/X = window.close(), Enable = setter
[DesktopConnectionModalContentV2.test.tsx](vscode-webview://0pb5ovl2ua2akv0mebbsm43h24toinf485c21nlbhogvdkmsdpk8/pearpass-app-browser-extension/src/action/containers/Modal/DesktopConnectionModalContent/DesktopConnectionModalContentV2.test.tsx) — body, closeOnOutsideClick=false, onClose injection vs fallback, retry, no-onRetry path
window.close() intent confirmed: per user's earlier spec ("If the user clicks on 'X' icon, just close the modal and the extension"). Added closeOnOutsideClick={false} on both NonSecureWarningV2 and DesktopConnectionModalContentV2 Dialogs to prevent accidental backdrop-click kills. ESC/X/Discard still close popup (intentional behavior change documented in tests).

Verification: lint clean across 7 files. 20/20 tests pass. TS noise (TS2786) pre-existing across repo, unaffected.

nickgr2 (approved):
All five concerns from the prior review look resolved at 6ecbd01:

1. **Step 2 dead-end** — Step2Dialog.tsx:74 passes `onCancel={() => setSyncErrorMessage(null)}` and SyncingFailedModal renders a 'Re-enter code' secondary button when `onCancel` is provided. Code state is preserved across the toggle.

2. **Lost upstream error context** — `syncErrorMessage: string | null` replaces the boolean. Step2Dialog.tsx:53-61 prefers `res.error` (string) and falls back to the generic copy; the catch path uses `extractErrorMessage(err)`. SyncingFailedModal.tsx renders the message in a tertiary-color caption below the body.

3. **Spacing tokens** — `gap-[var(--spacing8)]` / `gap-[var(--spacing24)]` / `p-[var(--spacing24)]` / `rounded-[var(--radius8)]` applied across SyncingFailedModal, NonSecureWarningV2, and DesktopConnectionModalContentV2 — matches existing usage in onboarding/StepIndicatorBar and other V2 components.

4. **Tests** — three new files cover body copy, button visibility/wiring, `closeOnOutsideClick={false}`, retry/cancel paths, and onClose injection vs. fallback. Snapshots included.

5. **window.close() intent** — confirmed as spec-driven; `closeOnOutsideClick={false}` added on both Dialogs is a sensible safeguard against accidental backdrop kills.

LGTM.

boris91 (approved):

denislupookov:
@alvaropaco attach video / pictures please

torkinos (changes requested):
@alvaropaco lets please add screenshot / recording

torkinos:
Screenshot 2026-04-30 at 13 02 46
Screenshot 2026-04-30 at 13 08 11

torkinos (approved):

PR #172, opened 2026-04-28
Home Screen V2

shavtvalishvili:
## Summary

Home Screen V2 redesign for the popup: rebuilds the home screen on the new design system with a sidebar, redesigned header, and updated modals.

## Changes

**New surfaces**
- `SidebarV2` with vault selector, categories, folders, and a footer (authenticator/settings/lock).
- `RecordListV2` with grouped time-period sections, `EmptyCollectionViewV2`, and `EmptyResultsViewV2`.
- `MainViewHeader` with breadcrumbs, multi-select toggle, and sort menu.
- `AppHeaderV2` with search, import button, sidebar collapse toggle, and add-item menu (Password, Authenticator gated under `UNSUPPORTED`, wired import).
- `LayoutWithSidebar` wrapper used by `RecordListV2`.
- `AppHeaderContext` for header-level state.

**Vault management**
- `VaultSelector` with create, switch, and per-row actions (rename, invite).
- `CreateOrEditVaultModalContentV2`.
- Protected-vault unlock handled at the call site via existing `VaultPasswordFormModalContent`.

**New modals**
- `ImportItemOrVaultModalContentV2` — share-link entry with paste-to-import.
- `ImportVaultPreviewModalContent` — previews the imported vault before navigating.
- `ShareVaultModalContentV2` — QR + invite link for sharing the active vault.
- `MoveFolderModalContentV2` rewritten to mirror desktop (count title, custom-folder chips only, only items list scrolls).
- `DeleteRecordsModalContentV2` for multi-select delete.

**Polish & fixes**
- Window chrome: 650x550 popup, rounded corners, dark border, 4px padding.
- Eliminated white corners by painting html/body with the dark theme background.
- Disabled rubber-band overscroll on the popup.
- Vault and Settings surfaces share a rounded bordered main surface with a dark header.
- Sidebar collapse toggle moved into the `AppHeaderV2` hamburger button; sidebar fully hidden when collapsed; toggle flips 180° in collapsed state.
- Multi-select flow with actions bar and tooltips kept inside the viewport.
- Right-click context menu for record rows.
- Fade gradient at the bottom of overflowed modal item lists (move + delete).
- Folders and vaults sorted alphabetically by default.
- Sort menu width matched to desktop (260px).

## Recordings/Screenshots

https://github.com/user-attachments/assets/8f26c5b7-e38c-4b1a-8ed5-c3825afae551

https://github.com/user-attachments/assets/f9395124-e338-45ce-b10b-e9979ed106cb


boris91 (dismissed):

giorgikh93:
@nickgr2 please review

nickgr2 (commented):
A few findings on this V2 home-screen rework. Most are scoped to styling discipline and one runtime safety issue around the import flow.

**Styling discipline**

- `src/action/app/App.jsx` hardcodes `border: '1px solid #3D403F'` for the V2 popup chrome. Per `REVIEW_GUIDELINES.md` (#2 Styling discipline), no hardcoded hex colors in new UI — pull this from a token (e.g. `theme.colors.colorBorderPrimary`) via `useTheme()`. The matching `borderRadius: rawTokens.radius8` already goes through tokens; the border color should too.
- `src/shared/containers/LayoutWithSidebar/LayoutWithSidebar.styles.ts` and `src/action/pages/SettingsV2/index.tsx` use `borderRadius: 6` / `rounded-[6px]` directly. The rest of V2 standardises on `rawTokens.radius8`. Either route this through a kit token (add a `radius6` if the design genuinely calls for 6px) or align with `radius8`.
- `src/action/containers/RecordListView/RecordRowContextMenu.tsx` hardcodes a multi-stop `rgba(8,10,5,...)` `MENU_BOX_SHADOW`. If the kit exposes a shadow token (kit components like `ContextMenu` already render with their own shadow) prefer that — otherwise consider whether this whole menu should be a `ContextMenu` instance instead of a hand-rolled portal so the shadow stays consistent across all menus.
- `iconDestructive = theme.colors.colorSurfaceDestructiveElevated` in the same file uses a *surface* token to colour the `TrashOutlined` icon. Surface tokens are designed for backgrounds — if a `colorTextDestructive` (or similar) exists it'll match the destructive `NavbarListItem` text.

**Import flow runtime safety** — `src/shared/containers/ImportItemOrVaultModalContentV2/index.tsx`

```ts
const getPlatformInfo = (): Promise<{ os: string; arch: string }> =>
  new Promise((resolve) => {
    chrome.runtime.sendMessage({ type: 'GET_PLATFORM_INFO' }, resolve)
  })
// ...
const platform = await getPlatformInfo()
await addDevice(`${platform.os} ${platform.arch}`)
```

If the background errors / disconnects, `chrome.runtime.sendMessage` invokes the callback with `undefined` and sets `chrome.runtime.lastError`. The promise then resolves with `undefined`, the next access throws, and pairing is left half-applied (vault paired locally but `addDevice` never ran, or runs with `"undefined undefined"` if anything coerces). At minimum: check `chrome.runtime.lastError` inside the callback and reject the promise; ideally also fall back to a sensible label. The `try/catch` in `handleLoadVault` will catch the throw, but the toast says “check invite code” which would be misleading for a platform-info failure.

**Test coverage** — `REVIEW_GUIDELINES.md` (#10) calls for tests on new logic; the surfaces wired up here have very little:

- `RecordListV2`, `MainViewHeader`, `MultiSelectActionsBar`, `RecordRowContextMenu`, `SidebarV2` (incl. `VaultSelector`, `FolderRow`), `AppHeaderContainer`, `EmptyCollectionViewV2`, `EmptyResultsViewV2`, `useScrollOverflow`, `groupRecordsByTimePeriod`, `getRecordSubtitle`, and `sortByName` (used in many new code paths) all ship without tests. `groupRecordsByTimePeriod` in particular has non-trivial date-bucketing branches that are easy to exercise from a unit test and easy to regress on. The existing `MoveFolderModalContentV2.test.tsx` is the only test file in the diff.

**Minor**

- `src/shared/containers/ShareVaultModalContentV2/index.tsx` uses `dangerouslySetInnerHTML={{ __html: qrSvg }}` for the QR. The input is the freshly-minted `data.publicKey` round-tripped through `generateQRCodeSVG`, so the trust boundary is the lib — please confirm the `pear-apps-utils-qr` output is sanitised (or a known-safe template). If it isn't, render via an `` alternative.
- `src/shared/hooks/useScrollOverflow.ts` calls `useLayoutEffect(() => {...}, deps)` with a dynamic `deps` argument. `react-hooks/exhaustive-deps` will warn, and behaviour depends on stable references at every callsite — the current callers pass `[records.length]` inline, which creates a new array each render but with identical contents, so React's shallow compare keeps it stable. Worth a comment explaining the contract, or change the API to accept a single primitive trigger.
- `src/action/containers/MasterPasswordPrompt/MasterPasswordPrompt.tsx` drops `minHeight: '100vh', minWidth: '100vw'`. Confirm the parent now always provides full viewport height for both v1 and v2 — `App.jsx` sets a fixed pixel height on the outer wrapper, but `MasterPasswordPrompt` sits inside that and now relies on `h-full`/`w-full`.
- `src/shared/containers/SidebarV2/SidebarV2.styles.ts`: `SIDEBAR_WIDTH_COLLAPSED = 0` plus `overflow: hidden` makes the collapsed sidebar fully invisible, so the `FOLDERS_CHEVRON_CENTER_SHIFT_PX` math (and the `transform: translateX(...)` in `index.tsx`) is dead code while the sidebar is collapsed. Either remove it or set a non-zero collapsed width if the chevron is supposed to remain visible.
- `src/shared/containers/SidebarV2/index.tsx` `FolderRow` passes `onContextMenu` to `NavbarListItem`. The kit forwards rest props onto `react-strict-dom`'s `html.button`, which currently types it as `any` — works, but please add a quick test to lock in the right-click menu behaviour so a future kit change can't silently drop it.
- `src/action/containers/RecordListView/index.tsx`: the ROW_RECORD_ID_ATTR re-targeting effect adds a global `contextmenu` capture listener whenever a menu is open. Right-clicks on items *outside* the record list (e.g. on the sidebar) now also have `preventDefault` called on them while the menu is open. That probably matches intent (you want the next right-click to either retarget to another row or close), but worth confirming.

**Nit**

- `src/action/containers/AppHeaderV2/AppHeaderV2.styles.ts` uses the new CSS `rotate` property. Fine for Chrome MV3, but note Firefox <72 and older WebViews don't support the standalone `rotate` longhand — `transform: rotate(...)` is portable if cross-browser support is in scope.
- `src/shared/containers/SidebarV2/index.tsx` initial `isSidebarCollapsed = true`: the popup opens with the sidebar hidden, which means a fresh user lands on V2 home and sees no folders/categories until they discover the hamburger. PR description matches this, but worth a UX sanity-check.

shavtvalishvili:
@nickgr2 thanks for the thorough review. Pushed fixes for the runtime safety issue and started on tests; quick rationale on the items I'm leaving as-is:

**Styling**
- `#3D403F` border in `App.jsx` — `colorBorderPrimary` is `#212814` (dark olive) and `colorBorderSecondary` is `#2C3618`; neither matches the `#3D403F` neutral gray we want here. There isn't a token for this shade, so the hex stays until one is added to the kit.
- `borderRadius: 6` in `LayoutWithSidebar` and `SettingsV2` — intentional, this matches the designs (the rest of V2 is `radius8`, but those two surfaces are spec'd at 6px).
- `MENU_BOX_SHADOW` and the hand-rolled portal in `RecordRowContextMenu` — the kit's `ContextMenu` is a click-to-toggle dropdown anchored to a trigger element (positions at `triggerRect.bottom + 5` / `triggerRect.right - menuWidth`). The row menu is a right-click context menu that has to open at arbitrary cursor coordinates with viewport-edge repositioning and retargeting between rows on a second right-click — different shape, can't be expressed via the kit component as it stands today. Happy to revisit if the kit grows a cursor-positioned variant.
- `colorSurfaceDestructiveElevated` for the `TrashOutlined` icon — there's no `colorTextDestructive` in the theme; the available destructive tokens are `colorSurfaceDestructiveElevated` (`#D13B3D`, the bright accent), `colorSurfaceDestructive` (`#421010`, the dark surface), and `colorSurfaceError` (`#D1383D`). The "elevated" variant is what the kit itself uses as the destructive accent (e.g. on `NavbarListItem variant="destructive"` text), so colouring the icon with the same value keeps it consistent.

**Runtime safety** — fixed. `getPlatformInfo` now checks `chrome.runtime.lastError`, validates the response shape, and resolves `null` on failure (logged via `logger.error`). Caller falls back to a `'unknown'` device label so `addDevice` always runs and pairing can't land half-applied. The misleading "check invite code" toast no longer fires for platform-info failures.

**Tests** — added `groupRecordsByTimePeriod.test.ts` (covers favorites/all split, time-bucket boundaries, `createdAt` sort key, missing-timestamp fallback, asc-direction reversal) and `getRecordSubtitle.test.ts`. Will continue rolling tests onto the other surfaces, but didn't want to block this PR on full coverage.

**Nit** — `AppHeaderV2.styles.ts` switched from `rotate: '...deg'` to `transform: 'rotate(...)'` for portability.

**Leaving as-is**
- `MasterPasswordPrompt` height — same wrapper as v1, where this works fine; the parent in `App.jsx` always provides the full viewport.
- `isSidebarCollapsed = true` default — by design (matches the spec; the hamburger is the discovery point).
- `SIDEBAR_WIDTH_COLLAPSED = 0` and `FOLDERS_CHEVRON_CENTER_SHIFT_PX` — math becomes inert while collapsed but is load-bearing during the expand/collapse transition; not actually dead.
- Global `contextmenu` capture listener while menu is open — intended, that's how we retarget right-clicks to a different row without flicker.
- `dangerouslySetInnerHTML` for QR — input is the freshly-minted public key passed through `pear-apps-utils-qr`'s `generateQRCodeSVG`; no user input crosses the trust boundary.
- `useScrollOverflow` dynamic deps — both current callsites pass `[records.length]` (a primitive in an inline array; React's shallow compare keeps it stable). Will tighten the API if a future caller needs richer triggers.

nickgr2 (commented):
Thanks — the runtime-safety hardening on `getPlatformInfo` (lastError + response shape check + null fallback + try/catch around `sendMessage`) reads clean, and the new `groupRecordsByTimePeriod` / `getRecordSubtitle` tests cover the buckets and the fallback chain well. `transform: rotate(...)` swap is fine.

The styling-discipline items from my prior review still apply and are the main thing holding this back from approval:

- `src/action/app/App.jsx` — `border: '1px solid #3D403F'` hardcoded.
- `src/shared/containers/LayoutWithSidebar/LayoutWithSidebar.styles.ts` and `src/action/pages/SettingsV2/index.tsx` — `borderRadius: 6` / `rounded-[6px]` literals; the rest of V2 uses `rawTokens.radius8`.
- `src/action/containers/RecordListView/RecordRowContextMenu.tsx` — `MENU_BOX_SHADOW` hardcoded `rgba(...)`, and `iconDestructive` uses a surface token (`colorSurfaceDestructiveElevated`) for icon color.

Per `REVIEW_GUIDELINES.md` #2, hardcoded hex/spacing literals in new UI need to go through `useTheme()` + `rawTokens` (or a kit shadow token). Happy to defer the broader test-coverage gap and the remaining minor/nit items I raised earlier to a follow-up — just the four token bypasses above are blocking.

`minor:` PR description says "unknown device label" but the actual fallback string is just `'unknown'` — works either way, just flagging the minor description/code mismatch.

shavtvalishvili:
@nickgr2 No more hardcodes, updated ui-kit and reused the tokens.

nickgr2 (approved):
Token swap looks good. Verified at pinned ui-kit commit `217a6bd` (lockfile bump in this push):

- `colorBorderTertiary` = `#3D403F` — matches the prior hardcoded `#3D403F` in `App.jsx`
- `colorTextDestructive` = `#D13B3D` — same hex as the previous `colorSurfaceDestructiveElevated` for `iconDestructive`, and semantically more correct for an icon color (text token vs surface token)
- `rawTokens.shadowMenu` — string matches the prior `MENU_BOX_SHADOW` literal exactly
- `rawTokens.radius6` (and `var(--radius6)` exposed via `tailwind-theme.css`) = 6px — matches `borderRadius: 6` and `rounded-[6px]`

All five replacements are visually identical to the literals they replace. Resolves the open feedback about kit tokens not yet existing for these values.

torkinos (approved):

boris91 (approved):

PR #173, opened 2026-04-28
[Browser-Extension] Loading screen

giorgikh93:
### Requirements

- loading screen v2
### Screenshots/Recordings

Screenshot 2026-04-28 at 12 19 23

### dependencies

---
  - https://app.asana.com/0/0/1214260735921380

### DIFF
--- a/src/shared/components/LoadingV2/index.tsx
+++ b/src/shared/components/LoadingV2/index.tsx
@@ -1,11 +1,9 @@
 import { t } from '@lingui/core/macro'
 import { Trans } from '@lingui/react/macro'
-import { useLingui } from '@lingui/react'
 import { PageHeader, Text, useTheme } from '@tetherto/pearpass-lib-ui-kit'
 import { HourglassBottom } from '@tetherto/pearpass-lib-ui-kit/icons'
 
 export const LoadingV2 = () => {
-  const { i18n } = useLingui()
   const { theme } = useTheme()
   const primary = theme.colors.colorTextPrimary
 
@@ -19,7 +17,7 @@ export const LoadingV2 = () => {
           <div className="mb-[7px] flex items-center">
             <HourglassBottom color={primary} height={24} width={24} />
           </div>
-          <PageHeader as="h1" title={i18n._(t`Just a moment...`)} />
+          <PageHeader as="h1" title={t`Just a moment...`} />
         </div>
         <Text color={theme.colors.colorTextSecondary}>
           <Trans>
--- a/src/shared/components/LoadingV2/index.tsx
+++ b/src/shared/components/LoadingV2/index.tsx
@@ -1,0 +1,32 @@
+import { t } from '@lingui/core/macro'
+import { Trans } from '@lingui/react/macro'
+import { useLingui } from '@lingui/react'
+import { PageHeader, Text, useTheme } from '@tetherto/pearpass-lib-ui-kit'
+import { HourglassBottom } from '@tetherto/pearpass-lib-ui-kit/icons'
+
+export const LoadingV2 = () => {
+  const { i18n } = useLingui()
+  const { theme } = useTheme()
+  const primary = theme.colors.colorTextPrimary
+
+  return (
+    <div
+      className="bg-surface-primary text-text-primary flex h-full min-h-0 w-full min-w-0 flex-col items-center justify-center p-6"
+      role="status"
+    >
+      <div className="mx-auto flex w-full flex-col items-center gap-[6px]">
+        <div className="flex items-center gap-[6px]">
+          <div className="mb-[7px] flex items-center">
+            <HourglassBottom color={primary} height={24} width={24} />
+          </div>
+          <PageHeader as="h1" title={i18n._(t`Just a moment...`)} />
+        </div>
+        <Text color={theme.colors.colorTextSecondary}>
+          <Trans>
+            Connecting to your local vault to sync your latest credentials.
+          </Trans>
+        </Text>
+      </div>
+    </div>
+  )
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -3838,7 +3838,7 @@
     "node_modules/@tetherto/pearpass-lib-ui-kit": {
       "name": "pearpass-lib-ui-react-native-components",
       "version": "0.0.40",
-      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#68be4492aae52bd58552ab1fba9a3a60546af22e",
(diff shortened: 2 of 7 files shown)


giorgikh93:
@nickgr2 please review PR

src/shared/components/LoadingV2/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <div
+      className="bg-surface-primary text-text-primary flex h-full min-h-0 w-full min-w-0 flex-col items-center justify-center p-6"
+      role="status"
+    >
+      <div className="mx-auto flex w-full flex-col items-center gap-[6px]">
+        <div className="flex items-center gap-[6px]">
+          <div className="mb-[7px] flex items-center">
+            <HourglassBottom color={primary} height={24} width={24} />
+          </div>
+          <PageHeader as="h1" title={i18n._(t`Just a moment...`)} />
```
ndolidzee: Do we need both `i18._` and `t`` ` ? 
I think `useLingui` hook is unnecessary

ndolidzee (changes requested):

ndolidzee (approved):

nickgr2 (changes requested):
A few findings against `AGENTS.md` for this v2 work.

### Location: new kit-based component in a legacy-only directory

`src/shared/components/LoadingV2/index.tsx` is the **first** file in `src/shared/components/` that imports from `@tetherto/pearpass-lib-ui-kit`. `AGENTS.md` is explicit that this directory is for legacy v1 components and must not grow:

> `src/shared/components/` — legacy v1 custom components (do not grow; see rules below).

Other shared kit-based components in this repo live under `src/shared/containers/` (e.g. `MoveFolderModalContentV2/`, `CreateOrEditVaultModalContentV2/`, `AddDeviceModalContent/`, `PairedDevicesModalContent/`). Please move this component there (or to a feature-specific location like `src/action/app/`, since `App.jsx` is its only call site).

### File name: no v1 sibling, so no `V2` suffix

Per `AGENTS.md`:

> No v1 equivalent exists (net-new feature, net-new component) → create the file with its natural name, **no `V2` suffix**.

There is no `Loading` v1 sibling in the tree (`LoadingOverlay` is an unrelated full-screen overlay primitive). The natural name `Loading` should be used instead of `LoadingV2`. This also means `App.jsx` would import `Loading` directly and the `isV2()` branch keeps the same shape.

### Spacing literals should use kit tokens

`AGENTS.md`: "if the value corresponds to a semantic design decision (spacing step, brand color, radius), it must come from a token."

- `gap-[6px]` → `gap-[var(--spacing6)]` (the token exists as `spacing6`)
- `p-6` → `p-[var(--spacing24)]` (Tailwind's `p-6` resolves to the same 24px, but the guideline calls for the kit token explicitly)
- `mb-[7px]` — there's no `spacing7` token; if this is a vertical-alignment fudge for the icon vs heading baseline, prefer a flex `items-baseline` / icon-size adjustment over the off-by-one literal, or at minimum a comment explaining what it compensates for.

### Tests

Test coverage for both branches in `App.test.jsx` and the structural tests for `LoadingV2` look reasonable. A couple of nits:

- `LoadingV2/index.test.tsx` asserts on Tailwind class names (`expect(root).toHaveClass('h-full', 'w-full', ...)`). That's brittle — every layout tweak rewrites the test. Consider asserting on accessible role/label and rendered text instead, which is what `REVIEW_GUIDELINES.md` calls "real behavior, not implementation."
- The `@lingui/core/macro` mock relies on the babel macro plugin rewriting source-side calls; the test mock for `Trans` targets `@lingui/react` (not `/macro`). Worth confirming the rendered `'Connecting to your local vault…'` assertion is hitting your mock and not the actual macro output, just so the test isn't passing for the wrong reason.

giorgikh93:
@nickgr2 addressed your suggestions, please review again!

nickgr2 (changes requested):
Thanks for the rework — most of the prior CR is addressed cleanly:

- `LoadingV2` → `Loading` ✓
- Spacing literals replaced with kit tokens (`gap-[var(--spacing6)]`, `p-[var(--spacing24)]`); the `mb-[7px]` icon fudge is gone, replaced with `items-baseline` + `inline-flex shrink-0` on the icon span. ✓
- Test no longer asserts on raw Tailwind class names; it uses `getByRole('status')` / `getByRole('heading', { level: 1, name: '...' })` instead. ✓
- The `@lingui/core/macro` mock has been dropped — correct, since `babel.config.js` already has `@lingui/babel-plugin-lingui-macro` and `t\`...\`` is compiled to a literal in tests. ✓

### Still outstanding: file location

`src/shared/components/Loading/index.tsx` is still the only file under `src/shared/components/` that imports from `@tetherto/pearpass-lib-ui-kit` (verified via grep at `eb31b8b`). `AGENTS.md` is explicit on this directory:

> `src/shared/components/` — **legacy v1 custom components** (do not grow; see rules below).

> 2. **All new UI goes through the kit.** Any new `.tsx`/`.jsx` file — suffixed or not — must import from `@tetherto/pearpass-lib-ui-kit`, not from `src/shared/components/`.
> 3. **Never add variants under `src/shared/components/`** […] That tree is legacy […]

A rename keeps the file in the legacy tree, which is what the rule was meant to prevent. The component has a single call site (`src/action/app/App.jsx`), so the most natural home is `src/action/app/Loading/` — colocated with its consumer. Alternatively `src/shared/containers/Loading/` matches the precedent set by the other kit-based shared UI (`MoveFolderModalContentV2`, `CreateOrEditVaultModalContentV2`, `AddDeviceModalContent`, `PairedDevicesModalContent`). Either is fine; what matters is that `src/shared/components/` doesn't grow with this change.

### Out of scope / nit

`src/action/pages/SettingsV2/content/YourVaultsContent/index.tsx` adds `cursor-pointer` + `selectable` to the "Other Vaults" list rows — unrelated to the loading-screen feature and not described in the PR title/description. Worth either splitting into a separate PR or noting in the description; otherwise it's harder to bisect later.

giorgikh93:
@nickgr2 Addressed your requests, please review!

nickgr2 (approved):
All previously requested changes are addressed:

- Component renamed to `Loading` and moved to `src/action/app/Loading/` (out of legacy `src/shared/components/`).
- Spacing literals replaced with kit tokens (`var(--spacing6)`, `var(--spacing24)`).
- Tests assert semantic UI (`getByRole('status')`, `getByRole('heading', { level: 1, ... })`) instead of class names.
- Unrelated `YourVaultsContent` change removed.

LGTM.

boris91 (approved):

torkinos (approved):

PR #174, opened 2026-04-28
Redesign adding records modal and new records details

torkinos:
### Requirements
Redesign adding records modal and new records details

### Screenshots/Recordings

https://github.com/user-attachments/assets/6939b32a-87f6-4dac-8281-641dfdbba240

https://github.com/user-attachments/assets/73f2a7d1-f276-47d5-9137-0aeee8251f5a

https://github.com/user-attachments/assets/e74b4aa8-fdab-45f1-9478-60ba05893aa1

---
  - https://app.asana.com/0/0/1214113487981796


torkinos:
@nickgr2 please review

nickgr2 (changes requested):
Solid foundation for the v2 add/edit modal redesign. The dispatch wrapper + `useCreateOrEditRecord` hook is a clean way to keep call sites version-agnostic, and extracting `PasswordGeneratorV2` into a shared body that both the Dialog and the iframe wrapper consume is a good cleanup. A few items to address before merge.

### `1st Word` / `2nd Word` labels are hardcoded English

`getWordLabel` in `src/shared/containers/PassPhrase/PassPhraseV2.tsx` builds ordinal labels like `1st Word`, `2nd Word`, `21st Word` and feeds them into `InputField label={...}`. Those labels are user-visible and aren't routed through Lingui, so they won't translate. Two options:

- Drop the ordinal and use a cardinal label, e.g. `label={t\`Word ${position}\`}` — simplest and translates cleanly.
- If the ordinal is design-mandated, use Lingui's ordinal plural, e.g. `<Plural ordinal value={position} one="#st Word" two="#nd Word" few="#rd Word" other="#th Word" />`.

This is the only blocker for me — see also items below.

### Hardcoded design tokens in the new shared `PasswordGeneratorV2`

`src/shared/containers/PasswordGeneratorV2/index.tsx` carries over `rounded-[12px]`, `min-h-[41px]`, and `gap-3` from the previous content-popup version. These are design-system spacing/radius literals that should go through tokens (e.g. `rounded-[var(--radius12)]`, `gap-[var(--spacing12)]`). Moving the body into a shared component is the right time to reach for tokens — the old call site is going away anyway. Same file, lines around 201, 258, 300.

### Folder selector exists only in the Login modal — is that intentional?

`CreateOrEditLoginModalContentV2` exposes a `ContextMenu` + `SelectField` for picking/creating folders. The other six V2 modals (Note, Wifi, Custom, Identity, CreditCard, PassPhrase) all accept `selectedFolder` and persist `folder: selectedFolder ?? initialRecord?.folder ?? ''`, but they never render a folder picker, so once a record is saved into a folder it's effectively pinned to that folder when edited. Is this per design (only Login records get folder management for now), or is the folder picker missing from the non-Login forms? If the former, fine — if the latter, please add the same folder UI to the other modals.

### `useCreateOrEditRecord` has no v1 fallback for `recordType === 'password'`

In v2, the hook opens `<GeneratePasswordModalContentV2 />`. In v1 it falls through to the bottom `navigate('createOrEditCategory', { params: { recordType: 'password' } })` branch — which doesn't open a generator modal at all. RecordList works around this by keeping the v1 `<PasswordGeneratorModalContent />` path inline (`if (item.type === 'password') { if (isV2()) ... else setModal(...) }`), so it's safe. AppHeaderContainer relies on the hook directly (`if (type === PASSWORD_TYPE) handleCreateOrEditRecord({ recordType: 'password' })`) — that container is gated by `isV2()` in `App.jsx`, so today this is dormant rather than broken. Two minor asks:

- Either move the v1 `<PasswordGeneratorModalContent />` branch into the hook so all callers stay symmetric, or add a comment in `useCreateOrEditRecord` noting that `recordType: 'password'` is v2-only and explaining why the v1 fallback intentionally degrades to navigate.
- While there: the v1 fallback drops `selectedFolder` and `isFavorite` (the original `AppHeaderContainer` v1 path forwarded both). Consumed today only by AppHeaderContainer (v2-only) and RecordList (which never sent folder/favorite), so no regression — but worth carrying through for completeness if the v1 branch is meant to stay reachable.

### minor: doubled-up click target in mode rows

In the shared `PasswordGeneratorV2`, each radio row is a `<div role="button" tabIndex={0} onClick={() => setMode(option.key)} onKeyDown=...>` wrapping a `<Radio ... onChange={() => setMode(option.key)} />`. Both fire the same setter, and the keyboard handler on the outer div duplicates what the Radio already does for built-in keys. The keyboard handler addition is an improvement over the prior version, but consider just rendering the `Radio` (or wrapping it in a `<label>`) so there's a single accessible control rather than two stacked ones.

### Looks good

- All new UI imports from `@tetherto/pearpass-lib-ui-kit` — no slips into legacy `src/shared/components/`.
- `testID` / `data-testid` coverage on dialogs, buttons, fields, and dynamic list items is thorough.
- `V2` suffix usage is correct: new files with v1 siblings get the suffix, net-new files (`MultiSlotInput`-based shared bodies) follow existing naming.
- Lingui usage in the new modals is clean (titles, placeholders, button labels, aria labels all wrapped).
- No new SVGs, no hex literals, no CSP / manifest changes.
- Dispatch via `CreateOrEditCategoryWrapper` is a nice consolidation point.

src/action/containers/Modal/CreateOrEditCategoryWrapper/CreateOrEditNoteModalContentV2/CreateOrEditNoteModalContentV2.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+          <Text variant="caption" color={theme.colors.colorTextSecondary}>
+            {t`Additional`}
+          </Text>
+
+          <MultiSlotInput
+            testID="createoredit-note-v2-comments-slot"
+            actions={
+              <Button
+                variant="tertiary"
```
denislupookov: could you change please to tertiaryAccent

torkinos:
@nickgr2, please validate fixes. Only #3 (folder selector only on Login) stayed unchanged. Intentional, mirrors the desktop V2 modal exactly

nickgr2:
Head is still at `8541b32` — same as my last review. No new commits have landed yet, so there's nothing new to validate. Push the fixes and re-request review.

torkinos:
@nickgr2, pushed changes

nickgr2 (approved):
Thanks for the rev — all five points addressed cleanly. A small nit on the PassPhrase modal but nothing blocking.

### Re-review checklist

- **Ordinal word labels** — `selectOrdinal` in `PassPhraseV2.tsx` is the right call, translates correctly across locales.
- **PasswordGeneratorV2 tokens** — `rounded-[var(--radius8)]` + `gap-[var(--spacing12)]` good. (Kit only ships radius 8/16/20/26 — `--radius8` is the right pick for the 12px source value, my earlier suggestion of `--radius12` was off.) The remaining `min-h-[41px]` literal is a precise control-height constraint with no matching token, so I'd leave it.
- **`useCreateOrEditRecord` v1 fallback** — comment explains the v2-only contract for `recordType: 'password'` clearly, and `folderParam`/`favoriteParam` now flow through both edit and create v1 navigate calls. Good.
- **Doubled mode-row controls** — outer `role="button"` + `tabIndex` + `onKeyDown` removed, comment documents that the inner `Radio` is the accessible control and the outer `onClick` just extends the visual hit area. Acceptable shape.
- **Folder selector scope** — confirmed intentional in your prior reply (matches desktop V2). No change needed.

### Minor: PassPhrase hidden-message remove icon is always rendered

In `CreateOrEditPassPhraseModalContentV2.tsx:263-279`, the `PasswordField`'s `rightSlot` unconditionally renders the `TrashOutlined` remove button. The sibling modals (Custom, CreditCard, Identity, Wifi) gate it on `canRemove = customFieldsList.length > 1`, hiding the trash icon when only one row exists. With the new default of `[{ type: 'note', note: '' }]`, PassPhrase users can remove the only row, leaving the message list empty until they re-click "Add Another Message". Not a bug — submit filters empty notes — but worth aligning with the other modals so behavior is uniform across record types. Drop in the same `canRemove` guard.

### Looks good

- Custom modal getting a top-level `Comment` (note) field plus a `Hidden Message` list mirrors the Login modal pattern — nice parity.
- PassPhrase promoting `note` into its own `MultiSlotInput` and converting custom fields to `PasswordField`/"Hidden Message" matches the rest of the lineup.
- "Add Another Message" buttons consolidated to `variant="tertiaryAccent"` across all V2 modals — consistent.
- Submit-time `.filter((f) => f.note?.trim().length)` keeps empty default rows from being persisted.

denislupookov (approved):

boris91 (approved):

ndolidzee (approved):

giorgikh93 (approved):

PR #175, opened 2026-04-29
Current vault exp

giorgikh93:
### Requirements

Header copy update: Change "Your Vaults" section text to: "Manage your vaults. Select the vault you want to apply changes to."
UI Constraint
Display action icons (Sharing and 3-dots menu) only for the currently active vault in both the navbar dropdown (left side) and in the settings → "your vaults" list
Hide all action icons for inactive vaults
Interaction: Ensure all inactive vault rows are fully clickable.
Navigation: Clicking an inactive vault must switch the active context to that vault
Experience: The vault switch must be seamless,  the user must remain on the current page/view without a reload or redirect, but the new "current vault" should be reflected to the selected one
Validation: Confirm the new active vault now displays the Sharing and 3-dots menu, while the previous vault hides them.
### Screenshots/Recordings

https://github.com/user-attachments/assets/949a8db1-c85f-456b-9070-a37cd66c543b

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3847,7 +3847,7 @@
     "node_modules/@tetherto/pearpass-lib-ui-kit": {
       "name": "pearpass-lib-ui-react-native-components",
       "version": "0.0.40",
-      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#217a6bdef25364f5964158d1e5f4b8488db337d7",
+      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#636989931c65a63d4dd429911facbf38d60d1fc2",
       "license": "Apache-2.0",
       "dependencies": {
         "react-strict-dom": "^0.0.55"
--- a/src/action/pages/SettingsV2/content/YourVaultsContent/index.tsx
+++ b/src/action/pages/SettingsV2/content/YourVaultsContent/index.tsx
@@ -31,6 +31,7 @@ import { CreateOrEditVaultModalContentV2 } from '../../../../../shared/container
 import { PairedDevicesModalContent } from '../../../../../shared/containers/PairedDevicesModalContent'
 import { useModal } from '../../../../../shared/context/ModalContext'
 import { sortByName } from '../../../../../shared/utils/sortByName'
+import { useVaultSwitch } from '../../../../../shared/hooks/useVaultSwitch'
 
 const VAULT_ICON_BG = { backgroundColor: 'rgba(176, 217, 68, 0.18)' }
 
@@ -40,7 +41,7 @@ export const YourVaultsContent = () => {
     closeModal: () => Promise<void>
   }
   const { theme } = useTheme()
-
+  const { switchVault } = useVaultSwitch()
   const { data: vault } = useVault()
   const { data: allVaults } = useVaults()
 
@@ -114,7 +115,7 @@ export const YourVaultsContent = () => {
           as="h1"
           testID="settings-vault"
           title={t`Your Vaults`}
-          subtitle={t`Manage your vaults, control access permissions, and take protective measures if needed.`}
+          subtitle={t`Manage your vaults. Select the vault you want to apply changes to.`}
         />
       </div>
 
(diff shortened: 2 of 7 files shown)


giorgikh93:
@nickgr2 please review

nickgr2 (changes requested):
**Findings**

1. **`withRoundedBorders={false}` is not a valid `ListItem` prop** — `src/action/pages/SettingsV2/content/YourVaultsContent/index.tsx:228`. The kit's prop is `withRoundedBottomBorders` (note the "Bottom"), see `node_modules/@tetherto/pearpass-lib-ui-kit/src/components/ListItem/ListItem.tsx:34`. The unknown prop is silently dropped at runtime. The visual outcome is currently masked by the parent's `overflow-hidden rounded-[8px]`, but the prop intent is not actually applied — and other usages in this repo correctly use `withRoundedBottomBorders` (e.g. `src/contentPopups/views/PasswordGeneratorV2/index.tsx:352, 378, 401`). Either rename or drop the prop.

2. **`useVaultSwitch` `useCallback` deps are missing `refetchVault`** — `src/shared/hooks/useVaultSwitch.tsx:23-67`. `refetch` from `useVault` is recreated on every render and is referenced inside the callback (lines 45, 57) but not listed in the dep array. Functionally inert today (the underlying `dispatch` is stable), but `react-hooks/exhaustive-deps` will flag this and any future change to `refetch`'s captured state could become a stale closure. Add it to deps.

3. **`minor:` Floating promise in YourVaultsContent click handler** — `onClick={() => switchVault(v)}` (`YourVaultsContent/index.tsx:226`) returns a promise that's discarded. The other call site uses `void switchVault(...)` (`VaultSelector.tsx:77`); use the same pattern here for consistency with `no-floating-promises`.

4. **`useVaultSwitch` rethrows after logging — neither caller handles the rejection** — `useVaultSwitch.tsx:59-62` logs and rethrows, but both callers (`VaultSelector.handleVaultClick`, `YourVaultsContent` row click) fire-and-forget the promise. The result is an "Uncaught (in promise)" console error on every failure path, on top of the explicit `logger.error`. Either swallow after logging or surface the failure to the user (e.g. a toast/notification). What's the intended UX when the password modal submit or `refetch` fails?

5. **Question — VaultSelector dropdown no longer auto-closes after switch.** Removing the `onClose` prop means `isVaultSelectorOpen` stays true after a vault switch from the sidebar. Was the previous "switch then close dropdown" behavior intentional, or is this a UX regression? If intentional, ignore.

giorgikh93:
@nickgr2 please reveiw, I've addressed your points.

1. Please fetch new dependencies, I assume you have stale @tetherto/pearpass-lib-ui-kit
2. refetchVault is coming from the useVault hook and it's not wrapped in useCallback - meaning that it's value changing constantly and may cause un-necessary re-renders, that's why it's ommited form the deps array;
3. addressed
4. addressed
5. It's expected - dropdown shouldn't be closed

nickgr2 (approved):
Re-review at 0df27291.

**Resolved**

1. `withRoundedBorders` — you were right, I had stale deps. The current `design-system-v2` `ListItem.tsx` declares `withRoundedBorders?: boolean` and applies `[styles.containerFlatTop, styles.containerFlatBottom]` when `false`, so the prop is valid. Apologies for the noise.
2. `refetchVault` deps — fair. `react-hooks/exhaustive-deps` isn't enabled in the shared eslint config, and `switchVault`'s identity isn't tracked downstream, so the practical impact is nil. Up to you.
3. Floating promise — `void switchVault(v)` matches `VaultSelector`. ✓
4. Rethrow / unhandled rejection — toast on the unprotected path covers the user-visible failure. ✓
5. Dropdown auto-close — confirmed intentional. ✓

**Minor**

- `useVaultSwitch.tsx:53-55`: the new `} catch (error) { throw error }` inside the `onSubmit` body is a no-op — without it the rejection propagates identically through `await`. I'd just drop the inner `try/catch` and keep the `finally`-equivalent `setIsLoading(false)`, e.g.:
  ```ts
  onSubmit={async (password: string) => {
    setIsLoading(true)
    try {
      await refetchVault(vault.id, { password })
      closeModal()
      await onSuccess()
    } finally {
      setIsLoading(false)
    }
  }}
  ```
  Same behavior, less noise. Not a blocker.

LGTM.

boris91 (approved):

hakaari (approved):

PR #176, opened 2026-04-29
Redesign passkey

ndolidzee:
### Requirements
Redesign passkey

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3847,7 +3847,7 @@
     "node_modules/@tetherto/pearpass-lib-ui-kit": {
       "name": "pearpass-lib-ui-react-native-components",
       "version": "0.0.40",
-      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#217a6bdef25364f5964158d1e5f4b8488db337d7",
+      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#636989931c65a63d4dd429911facbf38d60d1fc2",
       "license": "Apache-2.0",
       "dependencies": {
         "react-strict-dom": "^0.0.55"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "pearpass-app-browser-extension",
-  "version": "1.6.1",
+  "version": "2.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "pearpass-app-browser-extension",
-      "version": "1.6.1",
+      "version": "2.0.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@lingui/core": "5.1.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "pearpass-app-browser-extension",
   "private": true,
-  "version": "1.6.1",
+  "version": "2.0.0",
   "type": "module",
   "license": "Apache-2.0",
   "scripts": {
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 3,
   "name": "PearPass",
-  "version": "1.6.1",
+  "version": "2.0.0",
   "permissions": [
     "storage",
     "nativeMessaging",
--- a/src/action/app/App.jsx
+++ b/src/action/app/App.jsx
@@ -57,20 +57,31 @@ export const App = () => {
 
       observerRef.current?.disconnect()
 
+      const computeClampedHeight = (rawContentHeight) => {
+        const contentHeight = Math.ceil(rawContentHeight)
+        if (!contentHeight) return null
+        const frameHeight = Math.max(0, window.outerHeight - window.innerHeight)
+        return Math.min(
+          DYNAMIC_WINDOW_MAX_HEIGHT + frameHeight,
+          contentHeight + frameHeight
+        )
+      }
+
+      // Immediate first sync — don't wait for the async ResizeObserver to fire
+      // (during which the OS shows the window at its default opening size).
+      const initialHeight = computeClampedHeight(
+        el.getBoundingClientRect().height
+      )
+      if (initialHeight) {
+        chrome.windows.update?.(currentWindow.id, { height: initialHeight })
+      }
+
       const observer = new ResizeObserver((entries) => {
         for (const entry of entries) {
           const blockSize =
             entry.borderBoxSize?.[0]?.blockSize ?? entry.contentRect.height
-          const contentHeight = Math.ceil(blockSize)
-          if (!contentHeight) return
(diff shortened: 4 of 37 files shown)


torkinos:
@nickgr2 please review

nickgr2 (changes requested):
Review of `redesign-passkey` (PR 176, head `0fe25d2`).

## Correctness

**`SelectPasskeyV2.tsx` — `handleCancel` races with `window.close()` and can drop the cancel signal.**
```ts
const handleCancel = () => {
  chrome.tabs.sendMessage(parseInt(tabId), { ... GOT_PASSKEY, credential: null })
  window.close()
}
```
`chrome.tabs.sendMessage` is async; closing the popup synchronously on the next line means the message is fired-and-forgotten and frequently won't reach the content script — leaving the page-side passkey flow waiting forever. Sibling handlers (`handleRecordSelect`, `handleGetHardwarePasskey`) correctly wrap `window.close()` in `.finally()`. Same fix here:
```ts
chrome.tabs.sendMessage(parseInt(tabId, 10), { ... }).finally(() => window.close())
```

**`PasskeyContainerV2.tsx` — `isVaultChanging` is dead code.**
```ts
const { isVaultChanging } = (() => {
  // isVaultChanging is tracked via useGlobalLoading indirectly through the modal flow
  // We surface it via the dropdown disabled state when a modal is open
  return { isVaultChanging: false }
})()
useGlobalLoading({ isLoading: isVaultChanging })
```
The IIFE always returns `false`, so the `useGlobalLoading` call is a no-op and `VaultSwitcherDropdown` is never disabled during a vault switch. Either wire up the actual loading signal (e.g. local `useState` flipped around the `refetchVault` call in `handleVaultSelect` and threaded into `disabled` on the dropdown), or remove the IIFE + `useGlobalLoading` entirely. Don't ship the placeholder.

**`CreatePasskeyV2.tsx` — dead `useState`.**
```ts
const [, setSelectedRecord] = useState<RecordEntry | null>(null)
```
Only the setter is used (in `onVaultChange`), but the value is never read. Either drop the state and pass `() => {}` to `onVaultChange`, or actually use the selected record.

**`App.jsx` — `ResizeObserver` effect has an orphan-observer race.**
```jsx
chrome.windows.getCurrent?.((currentWindow) => {
  ...
  observer.observe(el)
  observerRef.current = observer
})
return () => {
  observerRef.current?.disconnect()
  observerRef.current = null
}
```
`getCurrent` is async. If `windowSize.dynamic` flips to `false` (or the component unmounts) before the callback fires, the cleanup runs against `null`, then the callback later creates a new observer that's never disconnected. Track a `cancelled` flag captured by the effect closure and bail in the callback before constructing the observer.

**`App.jsx` — early `return` inside `for...of` exits the whole callback.**
```js
for (const entry of entries) {
  ...
  if (!contentHeight) return
  ...
  chrome.windows.update?.(currentWindow.id, { height: clampedHeight })
}
```
`return` here skips remaining entries instead of just this one — should be `continue`. Today the observer is bound to a single element so it doesn't bite, but it's a footgun if anything else ever observes more than one node.

## Styling discipline (guideline §2)

New files mix raw pixel literals with the design tokens used elsewhere. `PasskeyLoginCreate/index.tsx` correctly uses `gap-[var(--spacing8)]`, `mt-[var(--spacing12)]`, etc. — please align the new sibling files:

- `PasskeyContainerV2.tsx` line 167: `gap-[24px]` and `p-[12px]` → `var(--spacing24)` / `var(--spacing12)`.
- `PasskeyPopupHeader/index.tsx` line 36: `px-[12px] py-[10px]` — same.
- `VaultSwitcherDropdown/index.tsx`: `gap: '8px'` (inline style), `gap-[10px]`, `px-[12px] py-[10px]`, `mt-1`, transition `'150ms'`.
- `SelectPasskeyV2.tsx`: `p-[4px]`, `pb-[4px]`, `gap-[16px]`, `gap-[8px]`.
- `CreatePasskeyV2.tsx`: `p-[4px]`, `gap-[24px]`, `gap-[6px]`, `gap-[12px]`.

## i18n (guideline §5)

`PasskeyPopupHeader/index.tsx`: `aria-label="Close"` is user-facing (screen readers) and isn't going through Lingui — wrap with `t\`Close\``.

## TypeScript correctness (guideline §6)

Lots of `as` assertions in new TS code where narrowing or proper types would do better. Examples:

- `CreatePasskeyV2.tsx`: `(records as RecordEntry[])`, `(record as { type?: string })?.type`, `(error as Error)?.message || error`.
- `PasskeyLoginCreate/index.tsx`: `(websites as string[]).map(...)`, `(payload as { record?: { id?: string } })?.record?.id`, `(error as Error)?.message`.
- `SelectPasskeyV2.tsx`: `(records as PasskeyRecord[]).filter(...)`, `(error as Error)?.message`.

For the records cast in particular, prefer to type the hook return or narrow with a runtime type guard. For the error cast, `error instanceof Error ? error.message : String(error)` is the safer pattern and is what `logger.error(..., error)` already accepts.

## Minor

- `parseInt(tabId)` without radix in `CreatePasskeyV2.tsx`, `PasskeyLoginCreate/index.tsx`, `SelectPasskeyV2.tsx` (5+ call sites). Add the radix `10` — and consider validating that `tabId` is set before parsing; current code happily passes `NaN` to `chrome.tabs.sendMessage`.
- `PasskeyLoginCreate/index.tsx` line 102: `console.error` — sibling files in this PR use the project `logger` (already imported elsewhere). Keep it consistent.
- `PasskeyContainerV2.tsx` `useEffect` at line 95 omits `refetchUserData`, `refetchVault`, `refetchRecords`, `refetchVaults`, `navigate`, `setModal`, `closeModal` from deps. The `hasInitialized` ref guards against re-runs in practice, but the lint rule will flag this — either add the deps or annotate why they're intentionally excluded.
- `VaultSwitcherDropdown/index.tsx`: `disabled` is forwarded to `Pressable` (an `html.div` under the hood), where it has no native effect — only the in-handler `if (!disabled)` check actually disables interaction. That's fine as long as the visual disabled state is conveyed elsewhere; just noting that the `disabled` prop on the trigger is decorative.
- `PasskeyLoginCreate/index.tsx` line 94: empty `onChange={() => {}}` on the read-only `Passkey` `InputField`. If `readOnly` is supported (it's used here), `onChange` shouldn't be required — check whether the noop can be dropped.

## Out of scope but worth flagging

PR description is empty (the entire template is unfilled). Please add at least the requirements / changes / testing-notes sections — for a 1100-line redesign the test plan in particular is important: dynamic-height popup behavior across OS frame quirks (the existing Wayland comment in `useWindowResize.js` warns of this), what happens when the window grows past `DYNAMIC_WINDOW_MAX_HEIGHT`, and what happens on rapid vault-switch interactions.

torkinos:
### Screenshots/Recordings


https://github.com/user-attachments/assets/afddd0f7-fc61-4895-8928-47fb815122b2

PR #179, opened 2026-04-30
Redesign records details

torkinos:
### Changes
Implement new Record Details forms and components for various record types

### Screenshots/Recordings

https://github.com/user-attachments/assets/97c9601e-5860-4738-a1f4-4449c0775d90

### DIFF
--- a/src/action/components/OtpCodeField/OtpCodeFieldV2.tsx
+++ b/src/action/components/OtpCodeField/OtpCodeFieldV2.tsx
@@ -1,0 +1,69 @@
+import React from 'react'
+
+import { t } from '@lingui/core/macro'
+import { formatOtpCode, OTP_TYPE, useOtp } from '@tetherto/pearpass-lib-vault'
+import type { OtpPublic } from '@tetherto/pearpass-lib-vault/src/types'
+import { Button, InputField } from '@tetherto/pearpass-lib-ui-kit'
+import { LockOutlined, Sync } from '@tetherto/pearpass-lib-ui-kit/icons'
+
+import { TimerBar } from '../../../shared/components/TimerBar'
+import { useCopyToClipboard } from '../../../shared/hooks/useCopyToClipboard'
+
+interface OtpCodeFieldV2Props {
+  recordId: string
+  otpPublic: OtpPublic
+  testID?: string
+}
+
+export const OtpCodeFieldV2 = ({
+  recordId,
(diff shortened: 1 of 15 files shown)


torkinos:
@nickgr2 please review

hakaari (approved):

boris91 (approved):

nickgr2 (approved):
Read the 16 new files across the V2 record-details flow. Read-only detail views, kit-based, behind `isV2()`. A few `minor:` items, none blocking.

**`minor:` Contradictory ARIA on the divider** — `RecordListV2/index.tsx`:
```tsx
<div ... role="separator" aria-hidden="true" />
```
`role="separator"` and `aria-hidden="true"` cancel each other out. For a purely visual rule, drop `role`. For a semantic separator, drop `aria-hidden`.

**`minor:` Missing dep in `RecordDetailsV2.tsx`**:
```tsx
useEffect(() => {
  if (!record) handleCollapse()
}, [record])
```
`handleCollapse` closes over `onClose`, `params.source`, and `navigate`. exhaustive-deps will flag it; if any of those change between mount and the record-vanishing tick, the effect navigates with the stale `source`. Wrap `handleCollapse` in `useCallback` and include it.

**`minor:` Cast that defeats typing** — `LoginDetailsFormV2.tsx`:
```tsx
otpPublic={
  initialRecord.otpPublic as Parameters<typeof OtpCodeFieldV2>[0]['otpPublic']
}
```
`OtpPublic` is exported from `@tetherto/pearpass-lib-vault/src/types` (already imported in `OtpCodeFieldV2`). Type `LoginRecord.otpPublic` as `OtpPublic | undefined` and the cast goes away.

**`minor:` Inline styles vs className mix** — `RecordDetailsV2.tsx` and `WifiPasswordQRCodeV2.tsx` flip between `className="flex flex-col gap-[var(--spacing16)]"` and `style={{ display: 'flex', flexDirection: 'column', gap: ... }}` within the same file. Pick one primary form per file for readability.

**`minor:` `dangerouslySetInnerHTML` for QR SVG** — `WifiPasswordQRCodeV2.tsx`. Same pattern as the v1 component, and the QR lib encodes SSID/password as bitmap data rather than interpolated text, so no XSS surface introduced. Worth pinning down once in a kit-level wrapper so future QR usages do not have to re-litigate the safety call at every site.

PR #206, opened 2026-05-08
Browser Extension: 2FA: Standalone 2FA Form

giorgikh93:
### Requirements

- Standalone 2FA Form
### Changes

- Added authenticator option in the dropdown
- Fixed the flow to rely on RECORD_TYPES.OTP in navigation
### Testing Notes

- tested locally
### Screenshots/Recordings

https://github.com/user-attachments/assets/4ff8422b-89ed-4d62-9b0f-3f8f1c9c128c

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5712,7 +5712,7 @@
     "node_modules/@tetherto/pearpass-lib-vault": {
       "name": "pearpass-lib-vault",
       "version": "0.0.85",
-      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-vault.git#21d29ed5ba6c2c1272df6017389c0192f3178581",
+      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-vault.git#538be529307326322ef02c672d272bdf9137fb88",
       "license": "Apache-2.0",
       "dependencies": {
         "@reduxjs/toolkit": "2.5.0",
--- a/src/action/containers/AddItemContextMenu/index.tsx
+++ b/src/action/containers/AddItemContextMenu/index.tsx
@@ -1,24 +1,19 @@
 import type { ComponentType, ReactNode, SVGProps } from 'react'
 
 import { t } from '@lingui/core/macro'
-import { UNSUPPORTED } from '@tetherto/pearpass-lib-constants'
+import { AUTHENTICATOR_ENABLED } from '@tetherto/pearpass-lib-constants'
 import {
   ContextMenu,
   NavbarListItem,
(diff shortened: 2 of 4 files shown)


giorgikh93:
@nickgr2 please review

nickgr2 (commented):
A couple of findings — overall looks like a clean refactor away from the `'authenticator'` string sentinel.

**Orphaned comment** — `src/action/hooks/useCreateOrEditRecord.tsx:21`

The deletion of `AUTHENTICATOR_TYPE` left half of its explanatory comment dangling:

```ts
// the LOGIN form with a narrowed field set (Title + OTP secret + Comments).
const PASSWORD_TYPE = 'password'
```

Either remove this line entirely or fold it into a comment near the new `RECORD_TYPES.OTP` flow so it still has context.

**Question — saved record type**

After this PR, picking "Authenticator Code" sets `recordType: RECORD_TYPES.OTP` in the menu, but `CreateOrEditLoginModalContentV2.onSubmit` still hardcodes `type: RECORD_TYPES.LOGIN` on save. So the user sees an OTP-flavored entry point but the record persists as `LOGIN` (with `data.otpInput` set). If `RECORD_TYPES.OTP` (newly added to the bumped vault pin) is intended to be a real record type, future code filtering by `record.type === RECORD_TYPES.OTP` won't find these. Is the plan to switch storage to `RECORD_TYPES.OTP` in a follow-up, or to keep these as LOGIN records and use `RECORD_TYPES.OTP` only as a menu sentinel? If the latter, worth a one-line comment near the new `SUPPORTED_V2_TYPES` entry so the next reader doesn't trip on it.

**minor:** `src/action/containers/AddItemContextMenu/index.tsx:10-13` — the new `@tetherto/pearpass-lib-vault` import (external) ends up after the relative internal imports, and the blank-line separator between external and internal groups is gone. Easy to land back into the external group above.

giorgikh93:
@nickgr2 please review, addressed your points

nickgr2 (approved):
All three points from the prior review are addressed:

- The dangling `// the LOGIN form with a narrowed field set...` comment is gone.
- The new comment block above `RECORD_TYPES.OTP` in `SUPPORTED_V2_TYPES` cleanly explains the menu-sentinel-vs-persisted-type relationship — answers the question and replaces the old context the deleted comment used to carry.
- `@tetherto/pearpass-lib-vault` import is back in the external group with the blank-line separator restored.

Looks good.

boris91 (approved):

torkinos (approved):

PR #208, opened 2026-05-11
2FA: Data Synchronization & Mapping

giorgikh93:
### Changes

- OTP remove flow: remove otp code and keep login record
- added sidebar to authenticator screen
- fixed navigation flickering issue when switching from 'authenticator' to 'all items'
### Testing Notes

- tested localy
### Screenshots/Recordings

https://github.com/user-attachments/assets/7280d58a-4889-4b80-a2dd-3e2631b17487

### DIFF
--- a/src/action/app/App.jsx
+++ b/src/action/app/App.jsx
@@ -1,5 +1,6 @@
 import { useEffect, useRef } from 'react'
 
+import { AUTHENTICATOR_ENABLED } from '@tetherto/pearpass-lib-constants'
 import { rawTokens, useTheme } from '@tetherto/pearpass-lib-ui-kit'
 
 import { useRedirect } from './hooks/useRedirect'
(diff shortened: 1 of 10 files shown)


giorgikh93:
@nickgr2 please review

nickgr2 (approved):
A couple of questions and minor items below — nothing blocking.

**Add Code refresh symmetry — likely existing, not new**
The empty-state "Add Code" button in `AuthenticatorView` wires up `onSaved` to sequentially refetch userData → masterVault → vault, but the AppHeader's `AddItemContextMenu` (now reachable on the authenticator page once the list is non-empty) calls `handleCreateOrEditRecord({ recordType: OTP, mode: 'authenticator' })` without an `onSaved` callback. Given `useRecords({ shouldSkip: true, ... })` in the view, will the visible list pick up the new code after adding via the header `+` menu? Worth confirming. If `useRecords` is subscribed regardless of `shouldSkip`, then the explicit refetch in `handleAddCode` is redundant; if it isn't, the AppHeader path needs the same refetch.

Minor: also consider running the three refetches in `Promise.all` rather than sequentially — there doesn't appear to be an ordering dependency, and the user is waiting on the modal.

**Unhandled rejection on strip-OTP failure**
In `useRecordActionItems.handleStripOtp`, the `onConfirm` callback re-throws after `setToast`. The modal's `handleDelete` does `await onConfirm()` inside a `try { ... } finally { ... }` with no `catch`, so the rejection bubbles up to the Button's `onClick` handler as an unhandled promise rejection. The user still sees the toast, but it's noisy in dev tools and risks future code (an error boundary, a logger) reacting to it.
Either catch in the modal's `handleDelete` for the `onConfirm` branch, or don't rethrow in `handleStripOtp` (only rethrow if you want the modal to stay open AND want the caller to know — but the modal already stays open by not calling `closeModal()`).

**Missing testID on the empty-state Add Code button**
`AuthenticatorView` line ~89 is a raw `<button>` without `testID`/`data-testid`. Per the kit guidelines, interactive elements should have one for E2E. The two icons elsewhere in this view also lack them but you're rewiring this click target — good moment to add one (e.g. `data-testid="authenticator-empty-add-code"`).

**Nit: defensive fallbacks unreachable in handleStripOtp**
```js
const { otpInput, otp, ...restData } = record?.data ?? {}
const { otpPublic, ...recordWithoutOtp } = record ?? {}
```
`handleStripOtp` is only reachable when `isAuthenticatorLoginRecord === true`, which already requires `record.type === LOGIN` and `record.otpPublic` to exist. The `?? {}` fallbacks can't trigger. Not worth a follow-up on its own, but if you're touching this code: dropping them makes the intent clearer.

**Tests**
Coverage on the strip path is solid — happy path, non-authenticator page, login-without-otp, source=authenticator on detail page, and the error toast. The index-based `result.current.actions[4]` is brittle (rest of the file uses the same pattern though, so consistent).

torkinos (approved):

boris91 (approved):

PR #209, opened 2026-05-11
[v2] [Extension] Redesign authenticator details screen , [v2] [Extension] Redesign authenticator screen

nikushaomanidze:
##### Changes

[v2] [Extension] Redesign authenticator details screen  
[v2] [Extension] Redesign authenticator screen

### V1/V2 note

This PR intentionally replaces the existing Authenticator view in-place instead of adding a separate `AuthenticatorViewV2` component. Unlike other pages that currently keep both v1/v2 variants behind `isV2()`, Authenticator did not have a parallel v2 counterpart, and the extension is already running with `EXTENSION_DESIGN_VERSION === 2`.

### Testing Notes

Tested on macOS Chrome.

Follow-up test coverage will be tracked separately for:
- empty state vs records state
- multi-select toggle
- TOTP/HOTP section split

https://github.com/user-attachments/assets/6362f5e9-30ed-420b-abb7-5dcc51c5abb3


boris91 (dismissed):

torkinos:
@nickgr2 please review

nickgr2 (changes requested):
A v2 redesign that mostly mirrors `RecordListV2` patterns and uses the kit correctly. A few things worth addressing before merge.

### Concerns

**1. v1 is fully removed at the same path (`src/action/pages/AuthenticatorView/index.jsx` → `index.tsx`)**
Other pages in `Routes.jsx` (`RecordList`/`RecordListV2`, `Settings`/`SettingsV2`, `CreatePasskey`/`CreatePasskeyV2`, etc.) keep both variants and gate via `isV2()`. The Authenticator now renders v2 unconditionally because `<AuthenticatorView />` is no longer behind `isV2()` in `Routes.jsx:86`. With `EXTENSION_DESIGN_VERSION === 2` this is fine in practice, but the asymmetry is surprising — either the v1 fallback should be preserved (e.g. add `AuthenticatorViewV2`), or the PR should call out that the Authenticator is the first page to drop v1 entirely.

**2. `@ts-expect-error` on `groupOtpRecords` (`AuthenticatorView/index.tsx:23`) is incorrect**
The package does declare it: `node_modules/@tetherto/pearpass-lib-vault/src/types.d.ts:16` exports a typed `groupOtpRecords` and an `OtpGroupResult` return type. Combined with `allowJs: true` and the JSDoc in `groupOtpRecords.js`, there is no error to suppress; the directive will itself fail TS as an unused-suppression. Remove it and use the package's `OtpGroupResult` instead of the `as unknown as { totpGroups: ...; hotpRecords: ... }` cast on lines 103–107.

**3. Type-cast smell on the modal handlers**
`selectedRecordObjects as unknown as Parameters<typeof MoveFolderModalContentV2>[0]['records']` (and the same for `DeleteRecordsModalContentV2`) bypasses a real type mismatch caused by the local `OtpRecord` shape. `src/shared/utils/groupRecordsByTimePeriod.ts:1` exports `VaultRecord` with an open index signature — both `RecordListV2` and the modals already accept it. Reusing `VaultRecord` (or extending it with the OTP-specific fields) lets you drop these `as unknown as` casts and the casts on `useAppHeaderContext`, `useCopyToClipboard`, `useFolders`, and `useRecords`. The review guidelines specifically flag `as unknown as` patterns.

**4. Hardcoded values where tokens exist**
- `AuthenticatorView/index.tsx:383` — `padding: '24px'` should be `${rawTokens.spacing24}px`.
- `AuthenticatorView/index.tsx:417` — `<strong style={{ fontWeight: 600 }}>` has no matching token; prefer a `Text` variant from the kit (`labelEmphasized` / `caption`) that already encodes the weight, rather than overriding inline.
- `AuthenticatorView/styles.ts:4` — `HEADER_MIN_HEIGHT = 44` is a magic number. No direct token, but worth a comment justifying it or using the closest token (`spacing40` / `spacing48`).
- `OtpCodeField/styles.ts:47,57` — `height: 6` and `minWidth: 28`: same — fine if intentional, but a one-liner constant with a name would document the intent.

**5. Inline style literals in render bodies**
Lines 295–304, 376–385, 449–455, 458–466, 467–474 build fresh style objects each render. Most are static. Move them into `createStyles` so the static styles aren't allocated per render (consistent with the rest of the file and with `RecordListV2`).

**6. `useTypedTimerAnimation` narrowing drops the `animated` parameter**
`OtpCodeFieldV2.tsx:17–20` types `useTimerAnimation` as a 2-arg function, but the hook accepts an optional third `animated` argument. The cast is also unnecessary if you use the JSDoc-inferred types — the package's JSDoc on `useTimerAnimation.js` already documents the return shape.

**7. Stale-state setter**
`AuthenticatorView/index.tsx:322` — `setIsMultiSelectOn(!isMultiSelectOn)` should be `setIsMultiSelectOn(prev => !prev)` to be consistent with the rest of the file and avoid closure footguns.

**8. No tests added for ~600 LoC of new UI**
Repo norm is ~60% coverage. At minimum, worth adding tests for: empty state vs records state, multi-select toggle, and the TOTP/HOTP section split.

### Minor

- `Breadcrumb items={[t\`Authenticator\`]}` is a single-item breadcrumb for a top-level page — consider whether `PageHeader` or `Title` from the kit is more appropriate here.
- `cardGrouped` in `OtpCodeField/styles.ts` adds `borderBottom` even when the OTP field is the last (or only) child of `MultiSlotInput`. `InputField` follows the same pattern via the kit, so the OTP card matches — just confirm visually that the last row in the credentials group doesn't end up with a double-line.

torkinos:
@nickgr2 please check again

Not addressed:
1 v1 fallback / isV2() gating: The original Authenticator view had no v2 counterpart to preserve (unlike RecordList/RecordListV2).Making this the first page to drop v1 entirely was intentional — there's no legacy code being removed, just a rewrite. We can add a note to the PR description calling this out explicitly if that helps with the asymmetry concern.

8 No tests: Acknowledged. Tracking separately — adding coverage for empty state, multi-select, and TOTP/HOTP split in a follow-up.

Minor: Single-item Breadcrumb: Noted. The Breadcrumb is consistent with how other top-level v2 pages are structured in this extension; we'll revisit if the kit adds a dedicated PageTitle component.

nickgr2:
On (1): fair — there was a v1 `index.jsx` being deleted, but no parallel `AuthenticatorViewV2` to gate. Replacing v1 entirely at this path is consistent with `EXTENSION_DESIGN_VERSION === 2` and the eventual removal of `isV2()`. A note in the PR description calling out the asymmetry would help anyone tracing the lack of a V2 sibling later.

On (8) and the Breadcrumb minor: ack.

I'll re-review the latest SHA in the next pass.

torkinos:
@nickgr2 updated

nickgr2 (approved):
All seven actionable concerns from the previous review are addressed:

- (2) Imports `OtpGroupResult` / `OtpPublic` from the package types and uses them directly; `@ts-expect-error` retained with an accurate comment (the directive applies because `groupOtpRecords` is exported from `src/index.js` but the package's `types.d.ts` isn't picked up via `package.json#types`). Acceptable.
- (3) The `as unknown as Parameters<...>['records']` casts on `MoveFolderModalContentV2` / `DeleteRecordsModalContentV2` are gone; `OtpRecord` now extends `VaultRecord`.
- (4) `padding: '24px'` → `${rawTokens.spacing24}px`; `<strong>` → `<Text variant="labelEmphasized">`; `HEADER_MIN_HEIGHT` now has a justifying comment.
- (5) Wrapper, `emptyState`, `detailsWrapper`, `separator`, and `detailsPane` inline objects moved into `createStyles`.
- (6) `useTypedTimerAnimation` cast now includes the optional `animated` parameter.
- (7) `setIsMultiSelectOn(!isMultiSelectOn)` → functional `(prev) => !prev`.
- (1) Acknowledged as intentional (full rewrite under `EXTENSION_DESIGN_VERSION === 2`, no parallel V2). (8) Tests called out as minor.

### Out of scope (not blocking, but worth splitting)

Commit b70a6221 mixes in unrelated CI changes that have nothing to do with the Authenticator redesign:

- Deletes `.github/workflows/build-browser-extension.yaml` (91 lines — removes the release/zip-artifact pipeline that fires on push to `main`/`dev`/`staging`).
- `.github/workflows/lint.yml` and `test.yml`: drops the `dev` branch trigger and adds `--legacy-peer-deps` to `npm install`.

The PR title/description don't mention CI. Deleting the build/release workflow is a significant repo-wide change that maintainers will likely want to evaluate separately — ideally split into its own PR so the Authenticator redesign isn't held up by (or quietly carries) a CI rework.

Otherwise this looks good.

boris91 (approved):

torkinos (approved):

PR #210, opened 2026-05-13
Adding workflows to validates the actual PR head SHA

kapildev421:
### Requirements

- Run a CI workflow (lint, unit tests) for every PR opened against `main` such that:
- The workflow validates the **PR head code**, not the target branch.                                                                                                                                               
-  Triggers on `opened`, `reopened`, and `synchronize`.
-  Runs only for trusted contributors (org members, collaborators, or external PRs explicitly opted-in by a maintainer via the `verify` label) — required because this is a public repo. 
- using scherermichael-oss/action-has-permission to check the permission

### Changes
 Replaces `lint.yml` + `test.yml` with a three-file CI pipeline

### Testing Notes

-   Open a fresh test PR with a small no-op change against `main`.
-   In the Actions tab, confirm a workflow run named **"PR CI"** appears within ~10s of opening.                                                                                                                     
-   Open the run → `authorize` job → "Check authorization" step. Confirm a `::notice::` line of the form `Same-repo PR — authorized` (for a branch on this repo) or `Trusted author (MEMBER) — authorized` (for a fork PR from an org member).                                                                                                                                                                                        
-   Open `ci / quality` job → "Checkout code" step. Confirm the `HEAD is now at <SHA>` line **matches the "Latest commit" SHA shown at the top of the PR** in the GitHub UI. This proves the workflow ran against the PR head, not `main`.                                                                                                                                                                                               
-   Confirm `Install dependencies`, `Run ESLint`, `Run tests` all complete green.
-   **Trigger `synchronize`**: push another commit (or `git commit --allow-empty -m "trigger sync" && git push`) to the PR branch → confirm a brand-new "PR CI" run starts automatically, and the previous run is cancelled by the concurrency group.                                                                                                                                                                                 
-   **Trigger `reopened`**: close the PR, then reopen it → confirm a brand-new "PR CI" run starts.                                                                                                                   
                                                                                            

### Things reviewers should pay attention to

- **No secrets passed to the reusable** — `ci-pr.yml` does not declare a `secrets:` block, so even if the PR head contained malicious test code, it has no token to exfiltrate.    

### Screenshots/Recordings
<img width="1403" height="792" alt="Screenshot 2026-05-13 at 10 35 33 AM" src="https://github.com/user-attachments/assets/8002ade2-22e1-4195-b986-31ac154061b8" />

### dependencies

- No dependent PRs in this repo


torkinos:
@nickgr2 please review

nickgr2 (changes requested):
Reviewing the new PR-head-validating CI pipeline. The security model in the new files is sound (authorize-then-checkout, no secrets passed to reusable, contents:read only, persist-credentials:false). A few findings below — main one is that the replacement isn't actually happening.

**Old workflows are still present (main finding)**

The PR description says "Replaces lint.yml + test.yml with a three-file CI pipeline", but neither file is deleted in this diff. Both still trigger on pull_request_target, and because neither passes a ref to actions/checkout, they check out the **base** branch (main) instead of the PR head — i.e. they lint/test main on every PR and never actually validate the PR's code. That is the bug this PR exists to fix.

After merge there will be four workflows running on every PR:
- lint.yml (validates main, green by definition for any PR)
- test.yml (same)
- ci-pr.yml (validates the PR head, gated by authorization)
- Push CI on the merge to main

The old two should be deleted in this PR — otherwise they keep producing false-positive green checks and obscure the new gate.

**Third-party action: scherermichael-oss/action-has-permission**

It's pinned by SHA (good), but the action is on 1.0.6 and hasn't been touched in years. Since it runs in a pull_request_target context with GITHUB_TOKEN, the supply-chain surface is real even though it's frozen by SHA. Two alternatives worth considering:
- Inline the permission check with a gh api repos/{owner}/{repo}/collaborators/{actor}/permission call in the same shell step that already exists.
- Use actions/github-script with collaborators.getPermissionLevel.

Either removes the third-party dep entirely with no behavior change.

**minor: labeled event fires on every label, not just 'verify'**

ci-pr.yml triggers on types [opened, reopened, synchronize, labeled]. Any label add (e.g. someone tagging the PR with "documentation") will spin up the authorize job. Cheap, but a one-line filter avoids it:

```yaml
jobs:
  authorize:
    if: github.event.action != 'labeled' || github.event.label.name == 'verify'
```

**minor: dead branch in authorize-pr/action.yml**

The shell check `[ "$EVENT" != "pull_request_target" ] && [ "$EVENT" != "pull_request" ]` covers pull_request, but the workflow only registers pull_request_target. Either drop that branch, or — if the intent is to make the composite reusable on pull_request workflows too — note it in the action description so future callers know.

**minor: branches scalar vs list**

ci-push.yml has `branches: main` while ci-pr.yml has `branches: [main]`. Both are valid YAML, just inconsistent.

**Out of scope, but worth flagging once**

Neither old nor new CI runs the build (`npm run build`) or any typecheck. The repo is TS + Vite with four build targets — lint and jest will not catch a broken vite config or a TS error in code that lint passes. Not a blocker for this PR, but mentioning so it stays on the radar.

kapildev421:
@nickgr2  please review

nickgr2 (approved):
Thanks — all five actionable findings from the prior review are addressed in this revision (delta `da715c5d…eb142a1`):

- `lint.yml` and `test.yml` deleted — the main finding (PRs no longer get false-positive green checks from workflows that validated `main`).
- `ci-pr.yml` authorize job now gated with `if: github.event.action != 'labeled' || github.event.label.name == 'verify'` — `labeled` runs only fire for the relevant label.
- `authorize-pr/action.yml` drops the `pull_request` fallback in the "non-PR event" branch, matching what the workflow actually registers.
- `ci-push.yml` `branches: [main]` is now consistent with `ci-pr.yml`.

The two items I flagged as non-blocking (third-party `scherermichael-oss/action-has-permission` supply-chain surface, and no build/typecheck step) are unchanged — leaving them for follow-up as noted.

Approving.

torkinos (approved):

hakaari (approved):

PR #211, opened 2026-05-14
Wire vault deletion UI and connect receive-side to action-exchange.

shavtvalishvili:
## Summary
Extension delete-vault UX wired end-to-end with the receive-side action handler and the new direct-delivery transport (merged in from `device-direct-actions`).

### UI
- Sidebar vault selector and Your-Vaults V2 settings page expose Delete. The V2 confirmation modal validates the master password, runs `deleteVaultLocal`, and falls back to a fresh unprotected "Personal" vault when the last vault is removed (matches `CreateOrEditVaultModalContentV2`).
- Sidebar vault row context menu gains "View Paired Devices" (switches the active vault, then opens `PairedDevicesModalContent`).
- Paired-device count pluralised (`1 other device` vs `N other devices`).

### Receive-side
- `useVaultAccessRevoked` subscribes to `pearpassVaultClient.on('vault-access-revoked', ...)` and runs the local delete via the existing Redux flow.
- Bails out when the vault is no longer in the local list (covers duplicate/self deliveries).

### Send-side
- "Erase from all devices" toggle calls `broadcastDeleteVault(vaultId)` before `deleteVaultLocal`.

### device-direct-actions (now merged)
- Cross-device delivery uses the personal-swarm direct transport with outbox retries; inbound envelopes are ed25519-signed and verified against the sender's master-vault writer pubkey. See lib-vault and lib-vault-core PRs.

### Misc
- Dev hook `__pearpassTriggerAccessRevoked` removed.
- Type declarations updated (src/tetherto-modules.d.ts): ACTION_TYPES, broadcastAction, `@tetherto/pearpass-lib-vault/src/instances`.

## Recordings

https://github.com/user-attachments/assets/4603885c-0c28-4e9f-9f13-bc2aeaf47de1

## Depends on
- https://github.com/tetherto/pearpass-app-desktop/pull/508
- https://github.com/tetherto/pearpass-lib-vault/pull/48
- https://github.com/tetherto/pearpass-lib-vault-core/pull/59
- https://github.com/tetherto/pearpass-lib-native-messaging-bridge/pull/11


torkinos:
@nickgr2 please review

nickgr2 (changes requested):
Reviewed `4ccbb6ff8c44b32bcb2bd3abfe6b178208dfc9ed`. One real bug in the fallback-vault-creation path, plus missing tests and a couple of smaller things.

## Bug — fallback `addDevice()` is called with no argument and always throws

`DeleteVaultModalContentV2/index.tsx:151` and `useVaultAccessRevoked.tsx:106`:

`useVault().addDevice` is typed as `(deviceName: string) => Promise<void>` (`src/tetherto-modules.d.ts:82`). Under the hood, the redux thunk passes `payload` straight into `addDeviceFactory(payload, vaultId)`, which throws `"Payload and vaultId are required"` when `payload` is falsy. Both new sites call `await addDevice()` with no arg, so after a successful `createVault({ name: 'Personal' })` the very next `addDevice()` always throws. The throw is caught locally and only logged — the user sees the toast `"A new \"Personal\" vault was created"` and is left with a vault that has no device registered.

Canonical pattern from `CreateOrEditVaultModalContentV2/index.tsx:128-132`:

```jsx
chrome.runtime.sendMessage(
  { type: 'GET_PLATFORM_INFO' },
  async (platform: { os: string; arch: string }) => {
    await addDevice(`${platform.os} ${platform.arch}`)
    ...
  }
)
```

Please apply the same pattern in both new code paths. Also worth checking why this didn't surface in development — there's no `type-check` script in `package.json`, so the missing required arg slipped past the build (strict mode would otherwise reject it).

## Missing tests for the new flows

Significant new code (`DeleteVaultModalContentV2`, `useVaultAccessRevoked`, `AccessRemovedModalContent`) lands with no colocated `*.test.tsx`. The pattern is well-established in this tree (`VaultSelector.test.tsx`, `useVaultSwitch.test.tsx`, `useRecordMenuItems.test.js`, etc.). The receive-side `useVaultAccessRevoked` is especially worth covering: it handles a remote-triggered event, performs a local delete, and may create a fallback vault — exactly the kind of cross-context message contract the review guidelines call out for at-minimum coverage. At a minimum I'd expect:

- `useVaultAccessRevoked`: payload missing `vaultId`, unknown `vaultId` (the new dedupe path from commit `56048b0`), active-vault delete with another vault remaining, active-vault delete with no remaining vaults.
- `DeleteVaultModalContentV2`: invalid master password (broadcast not called, no local delete), erase-all toggle on with broadcast failures (toast surfaces, local delete still runs), fallback `Personal` creation when last vault deleted.

## minor: "Erase Vault from N other devices with access" link replaces the open delete modal

`DeleteVaultModalContentV2/index.tsx:387-394` does `setModal(<PairedDevicesModalContent />)`, which discards the password and toggle the user has already entered. When the user dismisses the paired-devices modal, the delete flow is gone — they have to reopen it and re-enter the password. Either stack the modal (if the modal context supports it) or push the user back into the delete dialog on close. As-is the link is a footgun for anyone who clicks "let me double-check who that affects".

## minor: redundant required-check in `onSubmit`

`DeleteVaultModalContentV2/index.tsx:71-77` re-checks `!formValues.masterPassword` after the Validator schema already marks `masterPassword` required and the submit button is disabled when empty (`isSubmitDisabled = !values.masterPassword || isLoading`). The branch is unreachable — safe to drop.

## minor: `otherDeviceCount` is computed from `useVault().data.devices`, not the prop `vaultId`

`DeleteVaultModalContentV2/index.tsx:60-64` reads devices off the active vault. The two callers happen to ensure active vault === delete target (the V2 settings page only deletes the active vault; `VaultSelector.handleDelete` calls `switchVault(vault, () => setModal(...))` first). It works today, but the coupling between caller behaviour and modal correctness is non-obvious — a third caller could quietly show the wrong device count or call `addDevice` against the wrong vault on the fallback path. Worth either a code comment noting the invariant, or accepting `devices`/`otherDeviceCount` as a prop.

## Note on dependencies

Several APIs the PR consumes (`broadcastDeleteVault`, `ACTION_TYPES`, `deleteVaultLocal`, `pearpassVaultClient.on('vault-access-revoked', ...)`) live in unreleased lib-vault PRs per the PR description — I reviewed the contract shape against `src/tetherto-modules.d.ts` rather than installed sources. Worth landing those simultaneously so this branch doesn't merge to a broken state on `main`.

shavtvalishvili:
@nickgr2, I addressed:
- `addDevice()` fallback fix: new `initCurrentDeviceName` helper fetches OS+arch via the existing `platformMessages.getPlatformInfo()` bridge and calls `setCurrentDeviceName` on the vault lib; wired at all 3 `setPearpassVaultClient` sites.
- `tetherto-modules.d.ts`: `addDevice` now `() => Promise<void>`; `setCurrentDeviceName` declared.
- Tests added: `useVaultAccessRevoked`, `DeleteVaultModalContentV2`, `AccessRemovedModalContent`.
- Redundant `masterPassword` check removed.
- Toast on fallback Personal-vault failure.
- `latest.current` moved into `useEffect`.

I didn't address:
- "Erase Vault from N other devices" link left as-is: `ModalContext` already supports a modal stack (`setModal` pushes), so the link stacks on top of the delete dialog rather than replacing it.
- `otherDeviceCount` still read from active vault. Both callers ensure active === target before opening the modal; deferred as a follow-up to either pass devices as a prop or assert the invariant.

nickgr2 (approved):
All five points from the prior review are addressed:

- **Fallback `addDevice()` bug**: fixed via the `initCurrentDeviceName` indirection — platform info is fetched once at each `setPearpassVaultClient` site and stored on the vault lib via `setCurrentDeviceName`, so the no-arg `addDevice()` calls in the receive-side handler (`useVaultAccessRevoked.tsx:108`) and the last-vault fallback (`DeleteVaultModalContentV2/index.tsx:146`) can succeed. Different mechanism from the canonical `chrome.runtime.sendMessage` pattern I suggested, but equivalent in outcome and arguably cleaner.
- **Tests**: `useVaultAccessRevoked.test.tsx`, `DeleteVaultModalContentV2.test.tsx`, and `AccessRemovedModalContent/index.test.tsx` cover the cases I called out — including the dedupe path (`vault-access-revoked` with an unknown `vaultId`), fallback-Personal creation, broadcast-failure surfacing, and the access-removed lead string with and without `deviceName`.
- **Erase-from-N-devices link**: my prior concern here was incorrect. `ModalContext` (`src/shared/context/ModalContext.jsx:42`) pushes onto a stack rather than replacing, so the paired-devices view stacks on top of the open delete dialog and the user's password/toggle state is preserved when they dismiss it.
- **Redundant required-check**: removed.
- **Toast on fallback failure**: added in both `DeleteVaultModalContentV2/index.tsx:157` and `useVaultAccessRevoked.tsx:119`.

The `latest.current` assignment moving into a `useEffect` (`useVaultAccessRevoked.tsx:50`) is the right fix for the side-effect-in-render concern — the `useRef(...)` constructor seeds the ref so the first event can still find correct values before the effect runs.

## minor (follow-up)

`src/tetherto-modules.d.ts:82` narrows `addDevice` to `() => Promise<void>`, but seven existing call sites still pass a string and weren't updated:

- `src/action/pages/AddDevice/index.jsx:118`
- `src/action/pages/WelcomePage/CreateNewVault/index.jsx:62`
- `src/action/pages/WelcomePage/LoadVault/index.jsx:57`
- `src/shared/containers/CreateVaultModalContent/index.jsx:150`
- `src/shared/containers/CreateOrEditVaultModalContentV2/index.tsx:132`
- `src/shared/containers/ImportItemOrVaultModalContentV2/index.tsx:115`
- `src/action/containers/Modal/ImportItemOrVaultModalContent/index.tsx:106`

CI runs lint + tests but not `tsc` (`ci-reusable.yml` and no `type-check` script), so this won't surface, but the `.tsx` ones are strict-mode `TS2554`s. Behaviour stays correct (the new impl will use whatever `setCurrentDeviceName` stashed during `initCurrentDeviceName`, which happens to equal the string those sites pass), but the deviceName argument no longer does anything and the type contract is now dishonest. Worth a follow-up sweep to drop the argument at the call sites so the next reader doesn't assume it still flows through.

Depends-on PRs (lib-vault #48, lib-vault-core #59, native-messaging-bridge #11, desktop #508) still need to land in the same merge window — `setCurrentDeviceName`, `broadcastDeleteVault`, and the `vault-access-revoked` event aren't in the currently-installed `@tetherto/pearpass-lib-vault` yet.

giorgikh93 (approved):

PR #212, opened 2026-05-15
2FA + device name setup

denislupookov:
### Changes

2FA + device name setup

### Screenshots/Recordings

https://github.com/user-attachments/assets/c1692ec2-df64-49d2-8574-cfc77777b3df

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1225,9 +1225,9 @@
       }
     },
     "node_modules/@babel/plugin-transform-modules-systemjs": {
-      "version": "7.29.0",
-      "resolved": "https://registry.npmjs.org/@babel/plugin-transform-modules-systemjs/-/plugin-transform-modules-systemjs-7.29.0.tgz",
-      "integrity": "sha512-PrujnVFbOdUpw4UHiVwKvKRLMMic8+eC0CuNlxjsyZUiBjhFdPsewdXCkveh2KqBA9/waD0W1b4hXSOBQJezpQ==",
+      "version": "7.29.4",
+      "resolved": "https://registry.npmjs.org/@babel/plugin-transform-modules-systemjs/-/plugin-transform-modules-systemjs-7.29.4.tgz",
+      "integrity": "sha512-N7QmZ0xRZfjHOfZeQLJjwgX2zS9pdGHSVl/cjSGlo4dXMqvurfxXDMKY4RqEKzPozV78VMcd0lxyG13mlbKc4w==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -3008,9 +3008,9 @@
       }
     },
     "node_modules/@jest/diff-sequences": {
-      "version": "30.3.0",
-      "resolved": "https://registry.npmjs.org/@jest/diff-sequences/-/diff-sequences-30.3.0.tgz",
-      "integrity": "sha512-cG51MVnLq1ecVUaQ3fr6YuuAOitHK1S4WUJHnsPFE/quQr33ADUx1FfrTCpMCRxvy0Yr9BThKpDjSlcTi91tMA==",
+      "version": "30.4.0",
+      "resolved": "https://registry.npmjs.org/@jest/diff-sequences/-/diff-sequences-30.4.0.tgz",
+      "integrity": "sha512-zOpzlfUs45l6u7jm39qr87JCHUDsaeCtvL+kQe/Vn9jSnRB4/5IPXISm0h9I1vZW/o00Kn4UTJ2MOlhnUGwv3g==",
       "dev": true,
       "license": "MIT",
       "engines": {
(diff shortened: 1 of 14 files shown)


denislupookov:
@nickgr2 please review

boris91 (approved):

nickgr2 (changes requested):
A few concerns before this lands:

**1. `ModalContext.jsx` z-index drop from `10000` → `999` for v2 modals**

The wrapper for v2 modals is now `z-[999]`, which puts v2 modals **below** other top-level UI:
- `LoadingOverlay` is `z-[4000]` (`src/shared/components/LoadingOverlay/index.jsx:11`)
- `Toasts` is `z-[1500]` (`src/shared/components/Toasts/index.jsx:19`)
- `NonSecureWarning` is `z-[9999]` (`src/action/pages/NonSecureWarning/index.jsx:14`)

Before the change, v2 modals sat above all of these. After, a global loading state can mask an open modal entirely. The inner `z-[10000]` (line 103) doesn't help — once the outer wrapper creates a stacking context at 999, the inner z-index only competes within that context.

What's the intent here? If it's to let toasts/loading-overlay sit above a modal (intentional), please call that out in the description; if not, please leave it at `10000` or raise the wrapper to something above `9999`.

**2. `useRecordActionItems` — stripping OTP from a legacy LOGIN record on the Authenticator view is now data-destructive**

`AuthenticatorView` still surfaces records via `(records ?? []).filter((r) => r.otpPublic)` (`src/action/pages/AuthenticatorView/index.tsx:94`), which includes existing LOGIN records that carry `otpInput`. Before this PR, clicking *Delete element* on such a record from the authenticator page stripped only the OTP and left the rest of the login intact (the `isAuthenticatorLoginRecord ? handleStripOtp : handleDelete` branch). After this PR, that branch is gone and the same click deletes the whole login record — including the password.

If existing data has already been migrated and there are no legacy LOGIN+otp records left in user vaults, please describe that migration in the PR body. If migration hasn't happened, the strip-otp path needs to stay (or be moved to a confirmation copy that's explicit about deleting the entire record).

**3. `CreateOrEditCategoryWrapper` routes by `RECORD_TYPES.OTP`, but the new modal still creates `RECORD_TYPES.LOGIN`**

`CreateOrEditAuthenticatorModalContentV2` calls `createRecord({ type: RECORD_TYPES.LOGIN, … data: { otpInput, … } })` — i.e. the underlying storage shape is unchanged. The routing key `RECORD_TYPES.OTP` only exists as a UI discriminator. Two follow-ups:

- `RECORD_TYPES.OTP` is not defined in vault `0.0.85`. This PR bumps to `0.0.86`. Can you confirm `RECORD_TYPES.OTP` is exported by `0.0.86`? If it isn't, `recordType === RECORD_TYPES.OTP` reduces to `recordType === undefined`, which would route any caller that forgets to pass `recordType` into the authenticator modal.
- If `OTP` is purely a UI discriminator (not a stored record type), consider using a separate `mode: 'authenticator'` flag for routing rather than overloading `RECORD_TYPES`. The empty-state already passes `mode: 'authenticator'` to `handleCreateOrEditRecord` — routing by that would make the storage/UI split explicit.

**4. `AppWrapper/index.jsx` — module-level IIFE for `setPearpassVaultClient`**

```jsx
const client = new PearpassVaultClient({ debugMode: false })

;(async () => {
  const platform = await platformMessages.getPlatformInfo()
  const currentDeviceName = platform ? `${platform.os} ${platform.arch}`.trim() : undefined
  setPearpassVaultClient(client, { currentDeviceName })
})()
```

Two issues:

- The global `pearpassVaultClient` (in `@tetherto/pearpass-lib-vault/src/instances/index.js`) was registered synchronously before this PR. Now any consumer that touches it between module-load and promise resolution sees `undefined`. Has the consumer side been audited for this window?
- The IIFE has no `.catch(…)` — if `platformMessages.getPlatformInfo()` rejects (e.g. native messaging host not connected yet), the client is never registered at all and the rejection becomes an unhandled promise. Please attach a `.catch` that at minimum falls back to `setPearpassVaultClient(client)` so the app remains functional without device-name metadata.

The same pattern is duplicated in `src/shared/client.js` and `src/contentPopups/app/App.jsx`. Consider extracting it into a small helper (e.g. `setVaultClientWithDeviceName(client)`) rather than copy-pasting the 4-line block in three places.

**5. Missing tests for `CreateOrEditAuthenticatorModalContentV2`**

275 new lines, form + link-existing-login flow, no tests. Per `REVIEW_GUIDELINES.md` §10 this needs at least one component test covering the create path and the "link to existing login" path (the `updateRecords` branch is non-trivial).

**6. `CreateOrEditAuthenticatorModalContentV2.tsx` — type assertions**

The file leans heavily on `as string` / `Record<string, unknown>` casts (~25 sites). Two are particularly worth addressing:

- `as { data?: Array<{ id: string; data?: Record<string, unknown> }> }` on the `useRecords` return — define a typed `LoginRecord` interface and type the call instead. The vault library exports record types; if not, declare it in `src/tetherto-modules.d.ts` alongside the new export declarations you're already adding.
- `(formValues.otpSecret as string)?.trim()` — `useForm`'s `values` should be typed via the schema. Passing `Validator.object({ … })` to `useForm` typically gives you typed values; if it doesn't here, pass an explicit generic to `useForm`.

This isn't a blocker on its own but it's worth tightening while the code is new — the casts will be much harder to remove later.

**7. minor: Save button is `type="button"` and submits via `onClick={() => handleSubmit(onSubmit)()}`**

Because the `Form` has no `onSubmit` wired and the button isn't `type="submit"`, pressing Enter inside any field won't submit the form. Either set `type="submit"` and wire `<Form onSubmit={handleSubmit(onSubmit)}>`, or document the intent if Enter-to-submit should be disabled here.

**8. minor: `strict.css` adds `@layer reset;` with no comment**

Two-line file, no context for why the layer is being declared. A single-line comment ("ensures reset rules cascade below kit styles" or similar) would save the next reader a git blame.

torkinos (approved):

PR #222, opened 2026-05-28
feat: enhance copy to clipboard functionality across forms

ndolidzee:
### Screenshots/Recordings

https://github.com/user-attachments/assets/539e2a7b-f652-4073-ae2a-0a2af68f624f

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -91,12 +91,12 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.29.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.0.tgz",
-      "integrity": "sha512-9NhCeYjq9+3uxgdtp20LSiJXJvN0FeCtNGpJxuMFZ1Kv3cWUNb6DOhJwUvcVCzKGR66cw4njwM6hrJLqgOwbcw==",
+      "version": "7.29.7",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.7.tgz",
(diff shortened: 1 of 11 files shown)


ndolidzee:
@nickgr2 please review

nickgr2 (commented):
The change is small and consistent with the established `CopyButton` gating pattern in `src/shared/components/CopyButton/index.jsx`. Pulling `isCopyToClipboardEnabled` from the existing `useCopyToClipboard` hook is the right way to gate this UI — no kit violations, no new strings, no security regressions.

A few things to consider:

1. `package-lock.json` has ~2k lines of churn unrelated to this feature (Babel patch bumps, new `@react-native/*` peer entries, `dev:true` flag removals on existing entries). This looks like the result of running `npm install` rather than `npm ci`. `package.json` is unchanged vs `main`, so the lockfile diff is phantom churn — please regenerate via `npm ci` (or revert the lockfile chunk) to keep the diff focused on the feature.

2. Pre-existing in `useCopyToClipboard.js`, not introduced here, but worth being aware of as you wire the flag into more places: the hook captures the initial localStorage value with `useState(getIsCopyToClipboardEnabled())` and the `copyToClipboard` callback uses `useCallback(..., [])`, so a component that mounts before the user toggles the setting will keep showing/hiding copy affordances based on the stale value until it remounts. The detail forms typically remount on open, so this is probably fine in practice, but `AuthenticatorView` is more long-lived — if QA reports stale state after toggling, that hook is the place to fix it (move to a context, or subscribe to the storage key).

3. minor: in `AuthenticatorView/index.tsx` the inline cast `useCopyToClipboard() as { copyToClipboard: (text: string) => void; isCopyToClipboardEnabled: boolean }` is now wider; if you migrate `useCopyToClipboard` to `.ts` (it already has JSDoc types) the cast can go away entirely across the codebase.

PR #223, opened 2026-05-29
Add toast notification for disabled copy to clipboard action

ndolidzee:
### Screenshots/Recordings

https://github.com/user-attachments/assets/dea2de66-82fd-4194-adf9-56be34da21bb

### DIFF
--- a/src/shared/hooks/useCopyToClipboard.js
+++ b/src/shared/hooks/useCopyToClipboard.js
@@ -45,6 +45,9 @@ export const useCopyToClipboard = ({ onCopy } = {}) => {
 
   const copyToClipboard = React.useCallback((text) => {
     if (!isCopyToClipboardEnabled) {
+      setToast?.({
+        message: t`Please turn on the Copy to clipboard action from your settings`
+      })
       return false
     }
 


boris91 (approved):

nickgr2 (approved):
Reviewed the toast-on-disabled change in `useCopyToClipboard`. The behavior is sound: the string goes through Lingui (`t`), `setToast` is read from `useToast()` with optional chaining so it no-ops safely when no `ToastProvider` is mounted, and the early `return false` is preserved. Approving.

Two minor, non-blocking notes:

- `minor:` No test covers the new toast. The existing `copyToClipboard returns false when copy to clipboard is disabled` case asserts the return value and that `writeText` isn't called, but not that `setToast` fires. Since the hook reads `setToast` from context, asserting it would mean rendering the hook inside a `ToastProvider` (or injecting a spy) — worth adding so the new behavior is locked in.

- `minor:` The success toast passes `icon: Check`; this disabled toast has no icon. Intentional? If a warning/info icon exists in the kit it would make the two toasts visually consistent — not required.

Pre-existing (not introduced here, flagging for awareness only): `copyToClipboard` is wrapped in `useCallback(..., [])`, so the captured `isCopyToClipboardEnabled` / `setToast` are from first render. In practice each screen mounts its own hook instance and reads the flag from localStorage at mount, so it works — but if the setting is toggled and a copy is attempted within the same mounted instance, the stale closure would take the wrong branch. Out of scope for this PR.

giorgikh93 (approved):

PR #224, opened 2026-05-29
Replace InputField with TextArea for note input

ndolidzee:
### Screenshots/Recordings

https://github.com/user-attachments/assets/663d5b14-eb48-4f5a-bff9-92682ba975b5

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6032,7 +6032,7 @@
     "node_modules/@tetherto/pearpass-lib-ui-kit": {
       "name": "pearpass-lib-ui-react-native-components",
       "version": "0.0.40",
-      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#0301a142ff80a4c05b9efc245a9dc75a2bcdd8f2",
+      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#43e610d1b3a3c91168b02c8a0a5e8c4269d3e42d",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


ndolidzee:
@nickgr2 Please review, changes in package-lock are intentional and required for these changes, it updates lib-ui-kit which includes required changes for TextArea component

nickgr2 (approved):
Verified the swap from InputField to TextArea in both the create/edit form and the read-only details view.

- The kit bump (0301a14 -> 43e610d) adds `copyable` / `onCopy` to TextArea with `handleCopy = () => onCopy?.((rest.value as string) ?? '')`, mirroring InputField. So the copy button in NoteDetailsForm is preserved and `onCopy={copyToClipboard}` still receives the field value string. Confirmed by reading the kit diff at the new commit.
- The `onChange` signature change in CreateOrEditNoteModalContent is correct: TextArea exposes `onChange?: (value: string) => void`, so `(value) => noteField.onChange(value)` is right.
- i18n (`t`...``), testIDs, and validation are intact. Remaining changes are formatting-only.

minor: MultiSlotInput clones each child and injects `isGrouped: true` (`cloneElement(child, { isGrouped: true })`). TextArea doesn't declare an `isGrouped` prop, so it falls through `...rest` onto the native textarea, which will emit a React unknown-DOM-attribute warning and won't get the grouped border styling InputField applied. It's the sole slot here so it's cosmetic, but worth confirming the read-only note field still renders as intended.

Note: as a non-collaborator this approval won't satisfy branch protection.

boris91 (approved):

giorgikh93 (approved):

